PR #2, opened 2020-11-18
Stop basic functionality

leungjch:
This should be enough for stopping.
From https://docs.opencv.org/master/dd/d43/tutorial_py_video_display.html


Shrinjay:
lgtm!

PR #3, opened 2020-11-18
Added Class Functions

soumavm:
Added class functions for quit video feed, display video feed, and grab frame


Shrinjay:
Sick, congrats on the first commit!

PR #8, opened 2021-01-31
WIP: Gather geographical coordinates from pixel coordinates

Xierumeng:
DO NOT MERGE.

Function is untested, correctness unverified.

Unit tests must to be written for the function.

### DIFF
--- a/modules/geolocation/geolocation.py
+++ b/modules/geolocation/geolocation.py
@@ -10,11 +10,16 @@ class Geolocation:
         self.__cameraDirection3c = np.array([0.0, 0.0, -1.0])
         self.__cameraOrientation3u = np.array([1.0, 0.0, 0.0])
         self.__cameraOrientation3v = 1 * np.cross(self.__cameraDirection3c, self.__cameraOrientation3u)
-        self.__referencePixels = [[0, 0],
-                                  [0, 1000],
-                                  [1000, 0],
-                                  [1000, 1000]]
-        self.__cameraResolution = [1000, 1000]  # TODO Make global?
+        self.__referencePixels = np.array([[0, 0],
+                                           [0, 1000],
+                                           [1000, 0],
+                                           [1000, 1000]])
+        self.__cameraResolution = np.array([1000, 1000])  # TODO Make global?
+
+        self.__pixelsToGeographicals = np.array([[[0, 0], [0, 0]],
+                                                 [[1, 1], [1, 1]],
+                                                 [[2, 2], [2, 2]],
+                                                 [[3, 3], [3, 3]]])
 
         return
 
@@ -31,7 +36,7 @@ class Geolocation:
 
         pixels = np.array(self.__referencePixels).T
 
-        # Get u, v scaling from pixel coordinate and resolution
+        # Get scaling from pixel coordinate and resolution
         m = 2 * pixels[0] / self.__cameraResolution[0] - 1
         m = np.atleast_2d(m).T
         n = 2 * pixels[1] / self.__cameraResolution[1] - 1
@@ -48,7 +53,21 @@ class Geolocation:
 
 
         # Find intersection of the pixel line with the xy-plane
+        a = pixelsInWorldSpace.T
+        o = np.tile(self.__cameraOrigin3o, (points, 1))
+        o = o.T
 
+        # Verify that all pixels are pointing downwards
+        minimumZcomponent = 0.1  # This variable must be greater or equal to zero
+        # Replace invalid elements with false and check the resulting array for any false elements
+        a[2] = np.where(a[2] > minimumZcomponent, a[2], 0)
+        if (not np.all(a[2])):
+            return
 
+        # Formula
+        xCoordinates = o[0] - a[0] * o[2] / a[2]
+        yCoordinates = o[1] - a[1] * o[2] / a[2]
+
+        geographicalCoordinates = np.concatenate(xCoordinates, yCoordinates).T
 
         return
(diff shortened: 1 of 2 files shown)


modules/geolocation/geolocation.py
```diff
@@ -0,0 +1,68 @@
...
+
+        # Find corresponding geographical coordinate for every valid pixel
+        for i in range(0, np.size(self.__referencePixels)):
+
+            # Not enough pixels to create the map, abort
+            if (validPixelCount < minimumPixelCount):
+                return np.empty(shape=(0, 2, 2))
+
+            # Convert pixel to vector in world space
+            pixel = self.__referencePixels[i]
```
Shrinjay, line 48: Why not just directly access reference pixels? Not major but its a thing

modules/geolocation/geolocation.py
```diff
@@ -0,0 +1,68 @@
...
+            # Scaling in the u, v direction
+            m = 2 * pixel[0] / self.__cameraResolution[0] - 1
```
Shrinjay, line 50: Name these scale_factor_m or something, verbosity helps others understand this.

modules/geolocation/geolocation.py
```diff
@@ -0,0 +1,68 @@
...
+            n = 2 * pixel[1] / self.__cameraResolution[1] - 1
+
+            pixelInWorldSpace3a = c + m * self.__cameraOrientation3u + n * self.__cameraOrientation3v
```
Shrinjay, line 53: Where's c coming from? Would also create a better variable name like with scale_factor above.

modules/geolocation/unittests.py
```diff
@@ -0,0 +1,34 @@
...
+print(pair2)
+
+pairs = np.empty(shape=(0, 2, 2))
+print(pairs)
+pairs = np.concatenate((pairs, [pair1]))
+print(pairs)
+pairs = np.concatenate((pairs, [pair2]))
+print(pairs)
+pairs = np.concatenate((pairs, [pair3]))
+print(pairs)
```
Shrinjay, line 34: Let's talk about test cases on Thursday

modules/geolocation/geolocation.py
```diff
@@ -0,0 +1,68 @@
...
+
+        self.__pixelsToGeographicals = np.array([[[0, 0], [0, 0]],
+                                                 [[1, 1], [1, 1]],
+                                                 [[2, 2], [2, 2]],
+                                                 [[3, 3], [3, 3]]])
+
+        return
+
+    # TODO Placeholder, add functionality once we figure out how to convert raw plane data
+    def convert_input(self):
```
Shrinjay, line 27: Does this function call all functions required to convert camera points to geopoints? If not, then we need a function like that which is called by the parent after instantiation. Like the builder pattern, if you're familiar with that.

modules/geolocation/geolocation.py
```diff
@@ -0,0 +1,68 @@
+import numpy as np
+
+class Geolocation:
+    """Locates the geographical position of a set of pixels"""
+
+    # TODO Class members
+    def __init__(self):
```
Shrinjay, line 7: Create a constructor param for the array of points to convert to geocordinates, set pixelsToGeographicals (with new name) to that

modules/geolocation/geolocation.py
```diff
@@ -0,0 +1,68 @@
...
+        self.__cameraDirection3c = np.array([0.0, 0.0, -1.0])
+        self.__cameraOrientation3u = np.array([1.0, 0.0, 0.0])
+        self.__cameraOrientation3v = 1 * np.cross(self.__cameraDirection3c, self.__cameraOrientation3u)
+        self.__cameraResolution = np.array([1000, 1000])  # TODO Make global?
+        self.__referencePixels = np.array([[0, 0],
+                                           [0, 1000],
+                                           [1000, 0],
+                                           [1000, 1000]])
+
+        self.__pixelsToGeographicals = np.array([[[0, 0], [0, 0]],
```
Shrinjay, line 19: Are these the pixels we want to convert? If so, please give this a shorter name, I suggest pixelsToConvert

Shrinjay (changes requested):
Codebase looks good, well done, very impressed by your work as always. Made some nitpicky notes on code format and class shape. We'll talk about unit tests on Thursday.

modules/geolocation/geolocation.py
Xierumeng: Makes it clear that we're doing things on a single pixel. Using an alias or pointer was the intention here.

modules/geolocation/geolocation.py
Xierumeng: I thought we were re-using the class to save having to constantly do constructor and destructor calls? Also, pixelsToGeographicals will be populated by the output of the gather_point_pairs() function (currently a placeholder).

modules/geolocation/geolocation.py
Xierumeng: Output of gather_point_pairs() (currently has placeholder values).

modules/geolocation/geolocation.py
Xierumeng: This function converts raw plane data into something the module can do work on (see [documentation](https://docs.google.com/document/d/1VKYIrmWJYfpLPCAjQPiH1J5fnQPtNcjEC2NE5Eg4QxU/edit?usp=sharing)).

Xierumeng: I haven't yet thought of the name of the function that will run the entire thing from start to finish.

modules/geolocation/geolocation.py
Xierumeng: Mistake, should be replaced with "self.__cameraDirection3c".

PR #9, opened 2021-02-03
Search Program

Pie31415:


modules/search/search.py
```diff
@@ -0,0 +1,38 @@
...
+    angle : in
+      bearings notation in radians
+
+    Returns
+    -------
+    int
+      returns the minimum rotation angle
+
+    """
+    angle = math.radians(angle)
```
Shrinjay, line 24: Remain angle to heading, variable naming is important for self-documentation.

modules/search/search.py
```diff
@@ -0,0 +1,38 @@
...
+    # vector of the plane's current heading
+    v_heading = np.array([cos(angle), sin(angle)])
+    # vector from plane to tent
+    v_plane_tent = np.array(
```
Shrinjay, line 28: We're standardizing the codebase on camelCase instead of snake_case, so please change variables like v_heading to vHeading.

modules/search/search.py
```diff
@@ -0,0 +1,38 @@
+import math
+import numpy as np
+from numpy import sin, cos
+
```
Shrinjay, line 4: Do you intend for search.py to be a module? I think this should be a module since its processing logic, rather than process logic (i.e this is turning heading and GPS into a bearing, then the search **program** in the main process can take the bearing and send it off to the command module. 

Since this should be a module, put this inside a class, define an init function, make alt search a method of said class. Don't make any parameters of the function class member variables, since they're constantly regenerated they'll come in through a pipeline.

Shrinjay (changes requested):
Well done, just put logic inside a class and fix variable naming.

Shrinjay (approved):

PR #10, opened 2021-02-04
Video Mediator Class

AddeshB:
I feel like I might have missed out something, I'm not sure


modules/videoMediator/videoMediator.py
```diff
@@ -0,0 +1,20 @@
+import TargetAcquisition
+import DecklinkSRC
+
+class VideoMediator:
+    tent_coordinates = None
```
Shrinjay, line 5: Minor thing before merge, we standardize on camelCase for all variable names. Therefore target_acquisition should be targetAcquisition.

modules/videoMediator/videoMediator.py
```diff
@@ -0,0 +1,20 @@
...
+    decklinksrc=DecLinkSRC()
+    target_acquisition=TargetAcquisition()
+
+    def __init__(self):
+        self.__current_frame=None
```
Shrinjay, line 10: Thanks for adding this!

Shrinjay (changes requested):
Just change snake casing to camelCasing.

PR #13, opened 2021-02-05
Implemented main skeleton

Shrinjay:
Added a main skeleton function and program selection so subteams can start filling in their programs as well.


manumanuk (changes requested):
Documentation to CV standard needed

PR #15, opened 2021-02-11
Function to call train.py without conflicting main.py directory

AddeshB:


call train.py
```diff
@@ -0,0 +1,15 @@
+import os
+
+def _main_():
```
Shrinjay, line 3: Ok this is good, but can you add this as a function in the main.py file?

Shrinjay (changes requested):
Entire function is good, but please copy paste it into a function in the main.py file so I can merge.

PR #16, opened 2021-02-12
Geolocation gather coordinate pairs

Xierumeng:
Finds corresponding geographical coordinates from requested pixel coordinates, given camera intrinsics and extrinsics.


modules/geolocation/geolocation.py
```diff
@@ -0,0 +1,99 @@
...
+        """
+        Initial setup of class members
+
+        Returns
+        -------
+        Geolocation
+        """
+
+        # Input to gather_point_pairs()
+        self.__cameraOrigin3o = np.array([0.0, 0.0, 2.0])
```
Shrinjay, line 23: What's with these values for cameraorigin and cameradirection? As in why aren't they 0? Is it just a value so we can see where it should face?

modules/geolocation/geolocation.py
```diff
@@ -0,0 +1,99 @@
...
+        self.__cameraDirection3c = np.array([0.0, 0.0, -1.0])
+        self.__cameraOrientation3u = np.array([1.0, 0.0, 0.0])
+        self.__cameraOrientation3v = 1 * np.cross(self.__cameraDirection3c, self.__cameraOrientation3u)
+        self.__cameraResolution = np.array([1000, 1000])  # TODO Make global?
```
Shrinjay, line 27: Is this the actual value? If not, we need a TODO here.

modules/geolocation/geolocation.py
```diff
@@ -0,0 +1,99 @@
...
+        self.__referencePixels = np.array([[0, 0],
+                                           [0, 1000],
+                                           [1000, 0],
+                                           [1000, 1000]])
+
+        # Output of gather_point_pairs()
+        # Input to calculate_pixel_to_geo_mapping()
+        self.__pixelToGeoPairs = np.array([[[0, 0], [0, 0]],
```
Shrinjay, line 35: Are these numbers just placeholders?

modules/geolocation/geolocation.py
```diff
@@ -0,0 +1,99 @@
...
+        # np.array(shape=(n, 2, 2))
+        """
+
+        pixelGeoPairs = np.empty(shape=(0, 2, 2))
+        minimumPixelCount = 4  # Required for creating the map
+        validPixelCount = self.__referencePixels.shape[0]  # Current number of valid pixels (number of rows)
+        maximumZcomponent = -0.1  # This must be lesser than zero and determines if the pixel is pointing downwards
+
+        # Find corresponding geographical coordinate for every valid pixel
+        for i in range(0, self.__referencePixels.shape[0]):
```
Shrinjay, line 71: Why not just use validPixelCount self.__referencePixels.shape[0]

modules/geolocation/geolocation.py
```diff
@@ -0,0 +1,99 @@
...
+
+            # Not enough pixels to create the map, abort
+            if (validPixelCount < minimumPixelCount):
+                return np.empty(shape=(0, 2, 2))
```
Shrinjay, line 75: So is an empty 2x2 array always going to indicate failed processing? We need to be able to handle this.

Shrinjay (approved):
Approved, but there's a few questions I just want answered before I merge.

modules/geolocation/geolocation.py
Xierumeng: They're currently random defaults, I can set them to zero if you want.

modules/geolocation/geolocation.py
Xierumeng: Yes.

modules/geolocation/geolocation.py
Xierumeng: This is also a placeholder.

modules/geolocation/geolocation.py
Xierumeng: validPixelCount will be decremented if the current pixel in the loop is invalidated. If validPixelCount is used in the for logic then it will skip the pixels at the end.

modules/geolocation/geolocation.py
Xierumeng: Correct. Input verification would be handled in the main work function and the next function (calculate_mapping() in this case).

I would prefer to be able to use something like C++'s [std::optional](https://en.cppreference.com/w/cpp/utility/optional).

Also, it's an empty array of a pair of pairs.

modules/geolocation/geolocation.py
Shrinjay: Gotcha, leave it as it is then

modules/geolocation/geolocation.py
Shrinjay: Got it

modules/geolocation/geolocation.py
Shrinjay: Ah sorry, my bad for not understanding your architecture

PR #18, opened 2021-02-17
First pass at multithreading implementation 

Shrinjay:
Not for merge yet, just a first pass at implementing multithreading for the different components.


Xierumeng:
Prefix "WIP:" in the title to indicate that it should not be approved or merged yet.

pipelines/genericPipeline.py
```diff
@@ -0,0 +1,37 @@
+import threading
+from abc import ABC
+
+class GenericPipeline(ABC):
```
Xierumeng, line 4: Why not use [queue](https://docs.python.org/3/library/queue.html)?

pipelines/videoPipeline.py
```diff
@@ -0,0 +1,36 @@
+import threading
+from genericPipeline import GenericPipeline
+
+
+class VideoPipeline(GenericPipeline):
```
Xierumeng, line 5: This seems like a lot of copy-pasted code. Perhaps create a parameter `size` in the parent class constructor so that it can be generalized.

pipelines/coordinatePipeline.py
```diff
@@ -0,0 +1,8 @@
+from genericPipeline import GenericPipeline
+
+
+class CoordinatePipeline(GenericPipeline):
```
Xierumeng, line 4: Why does this exist? We can have two variables of the same type that are used for different things.

modules/decklinksrc/decklinksrc.py
```diff
@@ -1,40 +1,42 @@
...
 
-    def __init__(self):
+    def __init__(self, videoPipeline):
         self.__currentFrame = None
-        self.capture = cv2.VideoCapture('decklinkvideosrc mode=7 connection=0 ! videoconvert ! appsink') #Starts capture on initialization of object
-        #Because of this, we no longer need start stream code from Aryan
+        self.capture = cv2.VideoCapture(
+            'decklinkvideosrc mode=7 connection=0 ! videoconvert ! appsink')  # Starts capture on initialization of object
+        self.videoPipeline = videoPipeline
+        main = threading.Thread(target=self._main_)
```
Xierumeng, line 13: Improve naming (e.g. `mainThreadDeckLinkSRC`).

modules/decklinksrc/decklinksrc.py
```diff
@@ -1,40 +1,42 @@
...
 
-    def grab(self): #Logic for displaying single frame from deckLink
-        cv2.imshow('Grabbedframe', self.__currentFrame)
-    
-    def get_frame(self): #Logic for returning the current frame as a numpy array
+    def get_frame(self):  # Logic for returning the current frame as a numpy array
         return self.__currentFrame
-    
 
     def quitProgram(self):
```
Xierumeng, line 30: Is this supposed to be a destructor?

Xierumeng: Also possible duplicate of `stop()`.

modules/targetAcquisition/targetAcquisition.py
```diff
@@ -45,8 +46,17 @@ def __init__ (self, deckLinkFrame=np.zeros((416, 416, 3))):
         """
         self.boxes = [] # Contains BoundBox objects (see utils.py), each of which contains opposite corners of a rectangle by percentage of height and width of the image as (xmin, ymin) to (xmax, ymax)
         self.tentCoordinates = dict()
-        self.currentFrame = deckLinkFrame
-    
+        self.currentFrame = np.array()
+        self.videoPipeline = videoPipeline
+        self.coordinatePipeline = coordinatePipeline
+        main = threading.Thread(target=self._main_())
```
Xierumeng, line 52: Improve naming (see other comment). I was going to say to add a comment but if the name makes what's happening clear then it's a good name. Right now it is not a good name.

modules/videoMediator/videoMediator.py
```diff
@@ -1,20 +1,20 @@
...
-        initialize thread
-
-        """
-
+        self.videoPipeline = VideoPipeline()
+        self.coordinatePipeline = CoordinatePipeline()
+        self.coordinates = []
+        DeckLinkSRC(self.videoPipeline)
+        TargetAcquisition(self.videoPipeline, self.coordinatePipeline)
+        main = threading.Thread(target=self._main_)
```
Xierumeng, line 16: Improve naming (as above).

Xierumeng:
There also seem to be a couple of formatting issues and docstring issues.

pipelines/genericPipeline.py
Shrinjay: Threading is an accidental import here, don't need it

modules/decklinksrc/decklinksrc.py
Shrinjay: I have no idea, not my code but I'll clean it up

pipelines/videoPipeline.py
Shrinjay: The issue is that changing the pipeline to use a list in place of one package requires a change in add and get logic. That's why

pipelines/coordinatePipeline.py
Shrinjay: I'm not too sure about this design myself, I wanted to have an abstract class that all other pipelines could instantiate, but I took a lazy approach to it. I think I'm just going to not make generic abstract and instead make it the base pipeline and other modified pipelines can extend it

modules/decklinksrc/decklinksrc.py
Shrinjay: Got it

modules/targetAcquisition/targetAcquisition.py
Shrinjay: Yeah I made this mistake a lot, good catch

modules/videoMediator/videoMediator.py
Shrinjay: Also will do

Shrinjay (commented):
Got it, I'll review and make changes once Aryan does his reivew.

pipelines/genericPipeline.py
Xierumeng: Nothing to do with the threading library. What I mean is that the GenericPipeline class seems to re-implement the queue class.

pipelines/videoPipeline.py
Xierumeng: So it could possibly be generalized. A pipeline with a single element can be easily made with a list of size one.

pipelines/videoPipeline.py
```diff
@@ -0,0 +1,36 @@
...
+        """
+        Get oldest frame (front of array)
+        Returns
+        -------
+        frame
+        """
+        self.consumerLock.acquire()
+        currentFrame = self.package[0] if len(self.package) > 0 else None
+        self.producerLock.release()
+        return currentFrame
```
Xierumeng, line 35: The oldest element is not popped?

pipelines/videoPipeline.py
Xierumeng: Also the locking logic here will only ever allow a single element to be stored anyway.

modules/decklinksrc/decklinksrc.py
```diff
@@ -1,40 +1,42 @@
...
         self.capture.release()
         cv2.destroyAllWindows()
-    
 
-    def grab(self, frame): #Logic for grabbing frame from deckLink
-        self.cv2.imshow('Grabbedframe', frame)
-    
+    def grab(self):  # Logic for displaying single frame from deckLink
+        self.__currentFrame = self.capture.read()
+        return self.get_frame()
```
manumanuk, line 25: This function isn't really displaying the image like the old one was, it seems to just be retuning the video frame data

modules/decklinksrc/decklinksrc.py
```diff
@@ -1,40 +1,42 @@
...
         return self.__currentFrame
-    
 
     def quitProgram(self):
         self.capture.release()
         cv2.destroyAllWindows()
 
-
     # We need the file name, x dimensions, and y dimensions
     def recordVideo(self, filename, xdim, ydim):
```
manumanuk, line 35: This function doesn't seem to exit until the camera is no longer capturing any data - maybe it needs a new thread?

modules/targetAcquisition/targetAcquisition.py
```diff
@@ -34,7 +35,7 @@ class TargetAcquisition:
         Interprets BoundBox class values to populate self.tentCoordinates with centres of tents
     """
 
-    def __init__ (self, deckLinkFrame=np.zeros((416, 416, 3))):
+    def __init__ (self, videoPipeline, coordinatePipeline):
```
manumanuk, line 38: Please edit the documentation as appropriate

pipelines/videoPipeline.py
manumanuk: ^That seems more appropriate

pipelines/videoPipeline.py
```diff
@@ -0,0 +1,36 @@
...
+
```
manumanuk, line 36: Some implementation may need simplification/cleaning up or better documentation/UML update on architecture may be needed

manumanuk (commented):
Architecture seems to be getting pretty complex, should discuss a way to simplify it

Xierumeng:
I recommend rebasing rather than adding new commits when updating the branch.

pipelines/genericPipeline.py
Xierumeng: Class can be removed. From documentation:

> The Queue class in this module implements all the required locking semantics.

modules/decklinksrc/decklinksrc.py
manumanuk: This issue was fixed with the DeckLinkSRC merge (merge that in before final PR)

modules/decklinksrc/decklinksrc.py
manumanuk: I mean to say that the `recordVideo()` function is going to be constantly saving video, so the thread designated for this module will spend all its time saving video, up until the camera stream is shut off. We want it so that the video is saving simultaneously as other processes are happening (e.g. sending the video stream to another module). So I was saying maybe this video saver needs its own thread to enable that multitasking?

Shrinjay: You know what, I don't even think we need this functionality, I thought you were talking about quitProgram but frankly I don't see the value in refactoring something we're never gonna use anyways so I'll just drop it.

modules/decklinksrc/decklinksrc.py
Shrinjay: Done

PR #21, opened 2021-02-26
Updated Search.py

BaselOmari:
- Added an updated version for calculating the angle of rotation of plane.

- Changed return type from int to dict<"angle" : float, "counterclockwise" : bool>. This was done to instruct whether rotating counterclockwise would be more efficient (less degrees of rotation required).

- Fixed Typing Conventions and Formatting (includes filename change from 'search.py' -> 'Search.py')


modules/search/Search.py
```diff
@@ -0,0 +1,80 @@
...
+
+                "angle" : float
+                    Angle of rotation specified
+
+                "counterclockwise" : bool
+                    True if rotation should be counterclockwise
+                    False if rotation should be clockwise
+
+        """
+        import math
```
Shrinjay, line 34: Imports should always be at the top of the file

modules/search/Search.py
```diff
@@ -0,0 +1,80 @@
...
+
+        
+        '''
+            Calculating bearings between plane and tent
+        '''
+        planeLat = planeGPS["lat"]*math.pi/180
```
Shrinjay, line 40: Why are we converting lattitude into radians? Same with the tents.

modules/search/Search.py
```diff
@@ -0,0 +1,80 @@
...
+        planeLon = planeGPS["lon"]
+
+        tentLat = tentGPS["lat"]*math.pi/180
+        tentLon = tentGPS["lon"]
+
+        diffLon = (tentLon - planeLon)*math.pi/180
```
Shrinjay, line 46: This makes sense except for the *math.pi/180, same as above, I'm assuming this is a radian conversion?

modules/search/Search.py
```diff
@@ -0,0 +1,80 @@
...
+        elif diffBearing < 0:
+            if diffBearing < -180:
+                diffBearing = 360 + diffBearing
+            else:
+                diffBearing *= -1
+                cc = True
+
+        rotate = diffBearing
+
+        return {"angle": rotate, "counterclockwise": cc}
```
Shrinjay, line 80: I don't really like this interface, just have "direction" instead of "counterclockwise", or we could just give firmware a heading and let them figure out the most efficient route. I'd ask Anthony about that one.

Shrinjay (changes requested):
Overall your code is really good, I'm just a little confused by your approach. See comments for details.

modules/search/Search.py
BaselOmari: As mentioned in the previous comment, we need our latitude and longitude values to be in radians to work with the math library functions.

The values for the latitude of the tent and plane are used individually throughout the algorithm (which is why each one was converted into radians). 

However for the longitude values we only really care about the difference in longitude between that of the plane and the tent, thus we first calculated the difference and converted that into radians.

modules/search/Search.py
BaselOmari: Latitude and Longitude are given in degrees. The math.sin, math.cos, etc. functions all require inputs to be in radians, and thus we convert our values from the start to radians.

Looking back, I could have just used the math.radians() function to convert, however both methods perform the same process.

modules/search/Search.py
BaselOmari: Sure thing!

modules/search/Search.py
BaselOmari: I just thought it would be best if we calculated the most efficient route for them. However, I will contact Anthony and ask for his preference on this.

modules/geolocation/geolocation.py
```diff
@@ -129,3 +129,38 @@ def calculate_pixel_to_geo_mapping(self):
 
         # Return matrix product of mappedGeoMatrix and sourcePixelMatrixInverse
         return (mappedGeoMatrix.dot(sourcePixelMatrixInverse))
+
+    def map_location_from_pixel(self, pixels, coordinates):
```
Xierumeng, line 133: Please create a separate branch and pull request for the geolocation task.

In general each task should be separated into its own branch, and when ready for review, its own pull request.

PR #26, opened 2021-03-20
Map Location to Pixels Method for Geolocation

BaselOmari:


modules/geolocation/geolocation.py
```diff
@@ -129,3 +129,38 @@ def calculate_pixel_to_geo_mapping(self):
...
+            coordinates : np.array(shape=(5,2))
+
+        Returns
+        -------
+            np.array(shape=(5,2))
+        """
+        dehomogenized = np.array([])
+        
+        # Cycle through all 2D coordinates
+        for c in coordinates:
```
Xierumeng, line 149: My thought, rather than do a matrix multiplication for each vector individually, is to convert the input into a single 3xm matrix through [slicing](https://www.pythoninformer.com/python-libraries/numpy/index-and-slice/), the `.T` transpose function, and then extend the row and fill it with 1s. Once that is done, multiply this matrix by the transform.

On second thought, I do not know of any function that would let you dehomogenize the resulting matrix, and the output must be formatted back. The current for loop code may also be easier to read and understand.

modules/geolocation/geolocation.py
```diff
@@ -129,3 +129,38 @@ def calculate_pixel_to_geo_mapping(self):
 
         # Return matrix product of mappedGeoMatrix and sourcePixelMatrixInverse
         return (mappedGeoMatrix.dot(sourcePixelMatrixInverse))
+
+    def map_location_from_pixel(self, pixels, coordinates):
```
Xierumeng, line 133: `pixels` is not a good name. It's the transform matrix that is acting on the input pixels (which is currently called `coordinates`).

modules/geolocation/geolocation.py
```diff
@@ -129,3 +129,38 @@ def calculate_pixel_to_geo_mapping(self):
...
+    
+            # Compute Homogeneous Coordinates: Product of Image Pixels and Coordinates
+            homogeneousCoordinates = pixels.dot(c)
+
+            # Dehomogenizing the coordinate vector to compute the position in the destination image
+            dehomogenizedX = homogeneousCoordinates[0]/homogeneousCoordinates[1]
+            dehomogenizedY = homogeneousCoordinates[1]/homogeneousCoordinates[2]
+
+            
+            if dehomogenized.shape == (0,):
```
Xierumeng, line 161: Rather than doing this check, initializing like this: `dehomogenized = np.empty(shape=(0, 2))` would work better.

modules/geolocation/geolocation.py
```diff
@@ -129,3 +129,38 @@ def calculate_pixel_to_geo_mapping(self):
...
+        Parameters
+        -------
+            pixels : np.array(shape=(3,3))
+            coordinates : np.array(shape=(5,2))
+
+        Returns
+        -------
+            np.array(shape=(5,2))
+        """
+        dehomogenized = np.array([])
```
Xierumeng, line 146: Rename this to something better. Perhaps `geoCoordinates` given the input is currently `coordinates`?

modules/geolocation/geolocation.py
```diff
@@ -129,3 +129,38 @@ def calculate_pixel_to_geo_mapping(self):
...
+        # Cycle through all 2D coordinates
+        for c in coordinates:
+            # Express the 2D coordinates as 3D coordinates with the z value = 1 
+            c = np.concatenate((c, np.array([1])))
+    
+            # Compute Homogeneous Coordinates: Product of Image Pixels and Coordinates
+            homogeneousCoordinates = pixels.dot(c)
+
+            # Dehomogenizing the coordinate vector to compute the position in the destination image
+            dehomogenizedX = homogeneousCoordinates[0]/homogeneousCoordinates[1]
```
Xierumeng, line 157: Formatting: Add spaces around `/`.

Xierumeng:
Make sure to add unit tests as well.

modules/geolocation/geolocation.py
BaselOmari: Fixed! Changed 'pixels' to 'transformation_matrix' and 'coordinates' to 'pixels'.

modules/geolocation/geolocation.py
BaselOmari: Removed this check and changed initialization accordingly.

modules/geolocation/geolocation.py
BaselOmari: Renamed 'dehomogenized' to 'geoCoordinates'

modules/geolocation/geolocation.py
BaselOmari: Done!

modules/geolocation/geolocation.py
BaselOmari: Will get on this tomorrow. But for now, I have added `pixels = np.insert(pixels, 2, 1, axis = 1)` to convert all 2D coordinates of pixels to 3D coordinates with z = 1. 

However, the for loop still exists, and I will try and find a fix for that tomorrow hopefully.

BaselOmari:
Work done so far:
- Fixed Formatting (specifically division sign spacing)
- Renamed Parameters and Variables to more descriptive and appropriate alternatives
- Simplified Stacking of new dehomogenized coordinates

Work left:
- Add unit testing
- Figure out method to remove for loop and use slicing of arrays instead.

modules/geolocation/geolocation.py
```diff
@@ -129,3 +129,35 @@ def calculate_pixel_to_geo_mapping(self):
...
+            transformation_matrix : np.array(shape=(3,3))
+            pixels : np.array(shape=(5,2))
+
+        Returns
+        -------
+            np.array(shape=(5,2))
+        """
+
+        # Express all 2D coordinates of pixels as 3D coordinates with z value = 1
+        pixels = np.insert(pixels, 2, 1, axis = 1)
```
Shrinjay, line 148: Few things I'm not following on: 1) How come z=1? Are we not assuming that z=0 making the plane perpendicular to the horizon?

modules/geolocation/geolocation.py
```diff
@@ -129,3 +129,35 @@ def calculate_pixel_to_geo_mapping(self):
...
+        
+        # Cycle through all 2D coordinates of pixels
+        for p in pixels:    
+            # Compute Homogeneous Coordinates: Product of Image Pixels and Coordinates
+            homogeneousCoordinates = transformation_matrix.dot(p)
+
+            # Dehomogenizing the coordinate vector to compute the position in the destination image
+            dehomogenizedX = homogeneousCoordinates[0] / homogeneousCoordinates[1]
+            dehomogenizedY = homogeneousCoordinates[1] / homogeneousCoordinates[2]
+
```
Shrinjay, line 160: Where is this de-homogenizing equation defined? How is homogenousCoordinates a dot product, wouldn't transformation be achieved with a matrix-vector product which is not what np.dot does?

BaselOmari: I specifically followed steps 6 and 7 from this [post](https://math.stackexchange.com/questions/296794/finding-the-transform-matrix-from-4-projected-points-with-javascript).

As for the .dot() method, that can be used in numpy to perform matrix-vector multiplication. 

In `homogeneousCoordinates = transformation_matrix.dot(p)` I have multiplied the transformation matrix with the coordinates, to produce the homogeneous coordinates.

To be completely honest, I don't have a complete understanding of the math behind this, maybe Xierumeng can be of more help on that.

Xierumeng: According to documentation, `numpy.dot` is valid but it is preferred to use `numpy.matmul` or the `@` operator. See https://numpy.org/doc/stable/reference/generated/numpy.dot.html.

modules/geolocation/geolocation.py
Xierumeng: z=1 as the transformation is in R^3, but the input and output are in R^2. So it is required to convert the input to R^3 (this is called homogenization). After the transform, the result is also R^3 but needs to be converted to R^2 (dehomogenization). The reason this is done is so that we can use a linear (homogeneous) transformation (matrix multiplication) with a higher dimension instead of involving a matrix multiplication followed by an addition (a non-linear transformation).

See https://homepages.inf.ed.ac.uk/rbf/CVonline/LOCAL_COPIES/BEARDSLEY/node3.html (b) for an explanation (there was a better one than this for covering homogenization but I can't find it unfortunately).

modules/geolocation/geolocation.py
```diff
@@ -129,3 +129,40 @@ def calculate_pixel_to_geo_mapping(self):
 
         # Return matrix product of mappedGeoMatrix and sourcePixelMatrixInverse
         return (mappedGeoMatrix.dot(sourcePixelMatrixInverse))
+
+    def map_location_from_pixel(self, transformation_matrix, pixels):
```
Xierumeng, line 133: Formatting: camelCase variable names.

modules/geolocation/geolocation.py
Shrinjay: Ohhhhhhhhhhhhhhh right, my apologies, this was a really uninformed question lol I thought z=0 was from our plane thing. This is math 115 stuff I should've known, my apologies

modules/geolocation/geolocation.py
Shrinjay: Yeah please use numpy.matmul just to avoid confusion for future developers. As for the dehomogenization, I was just hoping to get some literature to verify its correctness, which you seem to have, so we're good.

Shrinjay (approved):
lgtm, but since I'm not final authority on the actual geolocation logic I'll leave the confirmation of that to Ryan. Seems like it homogenizes, transforms, then dehomogenizes which is exactly what we want. Good work :+1:

modules/geolocation/geolocation.py
```diff
@@ -129,3 +129,40 @@ def calculate_pixel_to_geo_mapping(self):
...
+
+        # Compute Homogeneous Coordinates: Product of Image Pixels and Coordinates
+        homogeneousCoordinates = np.matmul(transformationMatrix,pixels.T).T
+
+        geoCoordinates = np.empty(shape=(0, 2))
+        
+        # Cycle through all homogenized coordinates of pixels
+        for h in homogeneousCoordinates:
+            # Checking if the homogenized value of Z equals 0. If so, we return an empty array.
+            if h[2] == 0:
```
Xierumeng, line 158: I'm a bit worried about floating-point comparisons. Consider using something like `numpy.allclose()`. https://numpy.org/doc/stable/reference/generated/numpy.allclose.html

modules/geolocation/geolocation.py
```diff
@@ -129,3 +129,40 @@ def calculate_pixel_to_geo_mapping(self):
...
+                geoCoordinates = np.vstack((geoCoordinates, np.empty(shape = (2))))
```
Xierumeng, line 159: Use `float(‘inf’)` instead of empty array.

modules/geolocation/unittests.py
```diff
@@ -194,8 +194,68 @@ def test_point_set_2(self):
...
+        # Test
+        np.testing.assert_almost_equal(actual, expected)
+    
+    def test_transformation_matrix_1(self):
+
+        # Setup
+        transformationMatrix = np.array([[4,6,152],[120,5,99],[3,5,2]])
+        pixelCoordinates = np.array([[2,3],[12,99],[623,126],[1604,12],[0,4]])
+
+        expected = np.array([[7.739130435,15.39130435],
```
Xierumeng, line 228: Formatting nitpick: Try to keep it aligned.

modules/geolocation/unittests.py
```diff
@@ -194,8 +194,68 @@ def test_point_set_2(self):
...
+    def setUp(self):
+        
+        self.locator = geolocation.Geolocation()
+
+        return
+    
+    def test_identity_transformation_matrix(self):
+
+        # Setup
+        identityTransformationMatrix = np.ones(shape=(3,3))
```
Xierumeng, line 211: This is not an identity matrix. This is a matrix filled with 1s. Consider `numpy.eye()`.

Does this test pass? I'm looking at it and my current thought is no.

modules/geolocation/unittests.py
```diff
@@ -194,8 +194,68 @@ def test_point_set_2(self):
...
+        actual = self.locator.map_location_from_pixel(transformationMatrix, pixelCoordinates)
+
+        # Test
+        np.testing.assert_almost_equal(actual, expected)
+    
+ 
  
 if __name__ == "__main__":
-    unittest.main()
+    unittest.main()
```
Xierumeng, line 261: Formatting nitpick: Empty new line at end of file.

Xierumeng:
NOT APPROVED for merge until I specifically write it here.

modules/geolocation/unittests.py
BaselOmari: Done!

modules/geolocation/geolocation.py
BaselOmari: Done. I've replaced `np.empty(shape=(2))` with `np.full((2), np.inf)`.

modules/geolocation/geolocation.py
BaselOmari: Agreed. I have replaced the `np.testing.assert_almost_equal(actual, expected)` with `np.testing.assert_allclose(actual, expected, rtol = 1e-6)`. I set the relative tolerance of the assertion to 1e-6, which is a tolerance of about 0.0001%. All tests passed.

modules/geolocation/unittests.py
BaselOmari: I meant to say a transformation matrix filled with ones, instead of identity matrix.

Changed the name of the function and the associated variable name. Test Passed.

modules/geolocation/unittests.py
BaselOmari: Changed such that all rows of the matrices are on a new line:

`transformationMatrix = np.array([[5,1,5],
                                                       [9,1,2],
                                                       [0,0,0]])
`

modules/geolocation/geolocation.py
Xierumeng: No, what I meant was `if h[2] == 0:` on line 158. The unit tests were fine before.

BaselOmari: Sure thing. Changed `if h[2] == 0` to `np.allclose(h[2], 0)` and reverted tests back to `np.assert_almost_equal`.

Xierumeng:
Approved.

PR #31, opened 2021-03-24
Initialized Geolocation class at start of each Unit Test Class

BaselOmari:


modules/geolocation/unittests.py
```diff
@@ -182,20 +181,19 @@ def test_point_set_1(self):
...
-        actual = locator.calculate_pixel_to_geo_mapping()
+        actual = self.locator.calculate_pixel_to_geo_mapping()
         # Test
         np.testing.assert_almost_equal(actual, expected)
 
         
  
 if __name__ == "__main__":
-    unittest.main()
+    unittest.main()
```
Xierumeng, line 199: Formatting nitpicks: Try to keep the same formatting as the unit tests did before (in terms of line spacing).

Additionally, files should end with a new line without anything on it.

Other than that, approved.

PR #32, opened 2021-03-25
Implemented VideoMediator and multiprocessing in each module, cleaned up the code that makes it all work.

Shrinjay:


main.py
```diff
@@ -1,23 +1,21 @@
 import argparse
+import os
```
Shrinjay, line 2: Most of these changes are not related to the video mediator functionality, its just cleaning up this code/making it operable

tests/video_mediator_test.py
```diff
@@ -0,0 +1,9 @@
+import pytest
+from modules.videoMediator.videoMediator import VideoMediator
+
+
+# Integration test to check mediator behaviour
+def testVideoMediator():
```
Shrinjay, line 6: Don't think there's really much you can do for this module beyond integration testing

modules/decklinksrc/decklinksrc.py
```diff
@@ -28,28 +31,40 @@ class DeckLinkSRC:
...
+        mainProcess = mp.Process(target=self._mainProcess_, daemon=True)
+        mainProcess.start()
+
+    def _mainProcess_(self):
+        print("decklink started")
+        while True:
+            newFrame = self.grab()
+            if self.videoPipeline.empty() or newFrame != self.videoPipeline.get():
+                self.videoPipeline.put(newFrame)
+            sleep(0.1)
```
manumanuk, line 51: Any reason for the sleep?

modules/decklinksrc/decklinksrc.py
```diff
@@ -28,28 +31,40 @@ class DeckLinkSRC:
...
     def stop(self):
         """
         Logic for stopping video feed by releasing capture and destroying any windows open.
         """
         self.capture.release()
         cv2.destroyAllWindows()
-    
 
-    def grab(self, frame):
+    def grab(self):  # Logic for capturing single frame from deckLink
```
manumanuk, line 60: Why did you add this? It's repeated in the line below

modules/decklinksrc/decklinksrc.py
```diff
@@ -28,28 +31,40 @@ class DeckLinkSRC:
...
         """
         Logic for grabbing single frame from DeckLink
         """
-        cv2.imshow('Grabbedframe', self.__currentFrame)
-    
-    def get_frame(self):
+        self.__currentFrame = self.capture.read()
+        return self.get_frame()
+
+    def get_frame(self):  # Logic for returning the current frame as a numpy array
```
manumanuk, line 67: Same as above comment

modules/decklinksrc/decklinksrc.py
manumanuk: I'm also starting to wonder whether this function really serves any purpose; I can't think of many use cases where we would want to display a single frame

modules/targetAcquisition/targetAcquisition.py
```diff
@@ -43,10 +44,23 @@ def __init__ (self, deckLinkFrame=np.zeros((416, 416, 3))):
         deckLinkFrame : np.ndarray, optional
```
manumanuk, line 44: Please change the docs as appropriate whenever possible

modules/targetAcquisition/targetAcquisition.py
```diff
@@ -43,10 +44,23 @@ def __init__ (self, deckLinkFrame=np.zeros((416, 416, 3))):
...
             Variable size array containing data about a video frame (as given by cv2.imread())
         """
-        self.boxes = [] # Contains BoundBox objects (see utils.py), each of which contains opposite corners of a rectangle by percentage of height and width of the image as (xmin, ymin) to (xmax, ymax)
+        self.boxes = []  # Contains BoundBox objects (see utils.py), each of which contains opposite corners of a rectangle by percentage of height and width of the image as (xmin, ymin) to (xmax, ymax)
         self.tentCoordinates = dict()
-        self.currentFrame = deckLinkFrame
-    
+        self.currentFrame = np.zeros(2)
```
manumanuk, line 49: What's the purpose of initializing the frame to a size [2] array?

modules/targetAcquisition/targetAcquisition.py
```diff
@@ -75,17 +89,19 @@ def get_coordinates(self, newFrame=np.zeros((416, 416, 3))):
         # If new frame has been specified (is non-empty, set the current frame to the given frame)
         if np.count_nonzero(newFrame) != 0:
```
manumanuk, line 90: Don't know if this is the most effective implementation of this, now that I'm looking at it

manumanuk (commented):
-Still need to think about whether or not we need a process to be handling the video saving function (i.e. multiprocessing within the decklinksrc class to both allow for video saving and video processing, or combining these two features into the current implementation)

-Still need to implement a process for other classes (geolocation? command module?)

-Other than the above, no major flaws in actual implementation: props for following the styling guide so closely

modules/decklinksrc/decklinksrc.py
```diff
@@ -28,28 +31,40 @@ class DeckLinkSRC:
...
 
-    def __init__(self):
+    def __init__(self, videoPipeline):
         """
         Initializes DeckLink video stream
         """
         self.__currentFrame = None
-        self.capture = cv2.VideoCapture('decklinkvideosrc mode=7 connection=0 ! videoconvert ! appsink') #Starts capture on initialization of object
+        self.capture = cv2.VideoCapture(
+            'decklinkvideosrc mode=7 connection=0 ! videoconvert ! appsink')  # Starts capture on initialization of object
```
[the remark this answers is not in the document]
Shrinjay: I'll merge Ryan's PR to main, pull again and that should integrate the changes

modules/decklinksrc/decklinksrc.py
Shrinjay: Basically just so I don't flood the system, I'm not sure how good python is with memory management but I worry filling the pipeline too fast will cause issues

modules/targetAcquisition/targetAcquisition.py
Shrinjay: Oh lmao this was for testing, it should be an empty array

modules/targetAcquisition/targetAcquisition.py
Shrinjay: Ineffective in what sense?

manumanuk: Everytime this function runs, this piece of code will go through the entire 416x416x3 or whatever image to count the number of zeroes in order to determine if the current frame is valid

Might be a tiny difference in performance but nonetheless this isn't an ideal way to achieve what this line of code is trying to do

Shrinjay:
> -Still need to think about whether or not we need a process to be handling the video saving function (i.e. multiprocessing within the decklinksrc class to both allow for video saving and video processing, or combining these two features into the current implementation)
> 
> -Still need to implement a process for other classes (geolocation? command module?)
> 
> -Other than the above, no major flaws in actual implementation: props for following the styling guide so closely

To the first point, I don't think its a priority, the biggest use case for saving video is training, which we're already doing so the ship has kind of sailed there. Second point, once we hammer out this approach, I'm thinking of a way to generalize the whole multiprocessing creation process, maybe an abstract base class for all multiprocessed functions defining a default main process (thus forcing the creation of one by the user), I'll have tot hink through that

modules/targetAcquisition/targetAcquisition.py
Shrinjay: Performance wise its not an issue because numpy uses some voodoo parallelism under the hood, I suspect this function does boolean masking, but regardless when you use a built in numpy function to inspect an np array it typically doesn't iterate so performance is a non issue

modules/decklinksrc/decklinksrc.py
Shrinjay: Ok imo the get_frame function makes no sense at all, so I'll just take it out, none of the class attributes are private so no need for a getter.

modules/decklinksrc/decklinksrc.py
Shrinjay: This should just be to grab a single frame for processing, not display? The use case is grabbing a single frame for processing in targetAcquisition

main.py
```diff
@@ -28,7 +26,12 @@ def flightProgram():
         Send coordinates to command module
     Parameters: None
     """
-    return
+    print("start flight program")
+    videoFeed = VideoMediator()
+    # To get frame:
+    frame = videoFeed.coordinates.pop(0) if len(videoFeed.coordinates) != 0 else None
```
Xierumeng, line 32: Maybe an actual if statement with two returns.

modules/decklinksrc/decklinksrc.py
```diff
@@ -28,12 +31,23 @@ class DeckLinkSRC:
...
+        self.capture = cv2.VideoCapture(0)  # Starts capture on initialization of object
+        self.videoPipeline = videoPipeline
+        mainProcess = mp.Process(target=self._mainProcess_, daemon=True)
+        mainProcess.start()
+
+    def _mainProcess_(self):
+        print("decklink started")
+        while True:
+            newFrame = self.grab()
+            if self.videoPipeline.empty() or newFrame != self.videoPipeline.get():
```
Xierumeng, line 48: What is the intention here? This does not look like how a queue should be used.

modules/decklinksrc/decklinksrc.py
```diff
@@ -28,12 +31,23 @@ class DeckLinkSRC:
...
-        self.capture = cv2.VideoCapture(0) #Starts capture on initialization of object
+        self.capture = cv2.VideoCapture(0)  # Starts capture on initialization of object
+        self.videoPipeline = videoPipeline
+        mainProcess = mp.Process(target=self._mainProcess_, daemon=True)
+        mainProcess.start()
+
+    def _mainProcess_(self):
+        print("decklink started")
+        while True:
+            newFrame = self.grab()
```
Xierumeng, line 47: Is this guaranteed to return a valid frame? We do not want invalid data in the pipeline between processes. This is applicable to all pipelines.

tests/video_mediator_test.py
```diff
@@ -0,0 +1,9 @@
...
+    videoMediator = VideoMediator(testMode=True)
+    # By default if an empty frame is passed into the targetAcquisition class, "False" will be put into the pipeline
```
Xierumeng, line 8: NO.

tests/video_pipeline_test.py
```diff
@@ -0,0 +1,42 @@
...
+
+    for p in processes:
+        p.start()
+
+    assert coordinatePipeline.get() is None
+
+    exitSignal.put(True)
+
+
+def testPipelinePause():
```
Shrinjay, line 27: This test is kind of iffy tbh, it assumes that the qsize and pause happen almost instantaneously, which seems to be the case most of the time, maybe we can just get rid of it, or look at a way of checking when a process is blocked. Not too too high priority.

main.py
```diff
@@ -28,7 +29,21 @@ def flightProgram():
         Send coordinates to command module
     Parameters: None
     """
-    return
+    print("start flight program")
+    videoPipeline = mp.Queue()
```
Xierumeng, line 33: Add comment so we know between which two processes this is communicating.

modules/decklinksrc/decklinkSrcWorker.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def decklinkSrcWorker(pause, exitRequest, pipelineOut):
+    print("Start decklinksrc")
+    decklinkSrc = DeckLinkSRC()
+    while True:
+        pause.acquire()
+        pause.release()
+
+        curr_frame = decklinkSrc.grab()
```
Xierumeng, line 11: Is the method always correct? Will there always be a valid frame returned?

modules/targetAcquisition/targetAcquisitionWorker.py
```diff
@@ -0,0 +1,16 @@
...
+
+def targetAcquisitionWorker(pause, exitRequest, pipelineIn, pipelineOut):
+    print("start target acquisition")
+    targetAcquisition = TargetAcquisition()
+    while True:
+        pause.acquire()
+        pause.release()
+
+        curr_frame = pipelineIn.get()
+        coordinates = targetAcquisition.get_coordinates(curr_frame)
```
Xierumeng, line 12: Again, is this output always correct?

tests/video_pipeline_test.py
Xierumeng: Having race conditions is bad. We can reduce the chance though.

Consider this approach: pause, wait some time, get the size, wait some time, get the size and assert.

Unfortunately there aren't really any good ways of testing multiprocesses systems. In C++, Helgrind is one such system, but I don't know any for Python.

Xierumeng: It may not be worth testing multiprocess specifically because of non-determinism.

Shrinjay: Yeah, fuck it lol

modules/decklinksrc/decklinkSrcWorker.py
Shrinjay: No, I added a check to make sure its valid, however we could focus on source control, and make sure that invalid data is not put in the pipeline. Issue is, we do kind of want that invalid data there for testing.

modules/targetAcquisition/targetAcquisitionWorker.py
```diff
@@ -10,7 +10,8 @@ def targetAcquisitionWorker(pause, exitRequest, pipelineIn, pipelineOut):
 
         curr_frame = pipelineIn.get()
```
Xierumeng, line 11: There is no guarantee that the pipeline contains valid data, especially if dummy data needs to be inserted for a stop request. Of course, if we're okay with the stop request being literally us killing the process, this is moot (but still, correct code is good code).

modules/decklinksrc/decklinkSrcWorker.py
```diff
@@ -9,7 +9,8 @@ def decklinkSrcWorker(pause, exitRequest, pipelineOut):
         pause.release()
 
         curr_frame = decklinkSrc.grab()
-        pipelineOut.put(curr_frame)
+        if curr_frame is not None:
+            pipelineOut.put(curr_frame)
 
 
         if not exitRequest.empty():
```
Xierumeng, line 16: Add a comment. What does non-empty exitRequest mean? What is exitRequest at all?

modules/targetAcquisition/targetAcquisitionWorker.py
Shrinjay: I didn't consider the dummy data case, but this means we'll need to have validation once we pull data out, right? That's dependent on the shape of the data, including the dummy data. We need to define a) the shape of dummy data and b) use that to define the shape of valid data, and write a validator.

modules/decklinksrc/decklinkSrcWorker.py
Shrinjay: exitRequest is the exit pipeline. If its not empty, that means something was put into it and a stop order was issued.

modules/targetAcquisition/targetAcquisitionWorker.py
```diff
@@ -9,6 +9,10 @@ def targetAcquisitionWorker(pause, exitRequest, pipelineIn, pipelineOut):
         pause.release()
 
         curr_frame = pipelineIn.get()
+
+        if curr_frame is None:
+            continue
+
         coordinates = targetAcquisition.get_coordinates(curr_frame)
         if coordinates is not None:
```
Xierumeng, line 17: Make this consistent with above (i.e. if (condition): continue).

Xierumeng:
@Shrinjay Approved.

PR #34, opened 2021-03-26
Added Output Analysis calculation as a separate class

AddeshB:
Some functions are from the math library, so math is imported

### DIFF
--- a/modules/geolocation/geolocation.py
+++ b/modules/geolocation/geolocation.py
@@ -3,6 +3,7 @@ Geolocation module to map pixel coordinates to geographical coordinates
 """
 
 import numpy as np
+import math
 
 
 class Geolocation:
@@ -129,3 +130,71 @@ class Geolocation:
 
         # Return matrix product of mappedGeoMatrix and sourcePixelMatrixInverse
         return (mappedGeoMatrix.dot(sourcePixelMatrixInverse))
+    
+class OutputAnalysis:
+
+    def get_best_location(self,inputLocationTupleList):
+
+    # Splits the 3D numpy arrray into three separate arrays
+
+        try:
+            coordPair = np.vstack(inputLocationTupleList[:, 0]).astype(np.float64)
+            errorArray = np.vstack(inputLocationTupleList[:, 1]).astype(np.float64)
+            confidenceArray = np.vstack(inputLocationTupleList[:, 2]).astype(np.float64)
+
+    # Splits coordinate pair into x-coord and y-coord to remove outliers
+            xCoord = np.vstack(coordPair[:, [0][0]]).astype(np.float64)
+            yCoord = np.vstack(coordPair[:, [1][0]]).astype(np.float64)
+            xCoordTrimmed=[]
+            yCoordTrimmed=[]
+            errorTrimmed=[]
+
+
+            xCoordMedian = np.median(xCoord)
+            yCoordMedian = np.median(yCoord)
+            errorMedian = np.median(errorArray)
+
+    # A mmodified version of the trimmed mean calculation. Here the distance measuremnt is
+    # distance between every x,y point from the median coordinate pair, x median and y median.
+
+            for (x,y) in zip(xCoord,yCoord):
+                distance = math.sqrt((pow((x-xCoordMedian),2)+(pow((y-yCoordMedian),2))))
+
+                if distance<3.5:
+                    xCoordTrimmed.append(x)
+                    yCoordTrimmed.append(y)
+
+    # To handle the case where 2 points are too far apart from each other. Defaults to zero for both x and y
+
+                elif (distance>=3.5) and np.size(coordPair,0)==2:
+                    print ("Warning, only 2 data points and they are too far apart")
+                    xCoordTrimmed.append(0)
+                    yCoordTrimmed.append(0)
+                    break
+
+            for e in errorArray:
+                distance = math.sqrt((pow((e-errorMedian),2)))
+
+                if distance<3.5:
+                    errorTrimmed.append(e)
+
+
+    # Finds the average of the trimmed arrays using built-in np.average()
+
+            averageX = np.average(xCoordTrimmed, axis=0)
+            averageY = np.average(yCoordTrimmed, axis=0)
+            averageError = np.average(errorTrimmed, axis=0)
+
+
+            averagePair = (averageX,averageY)
+
+            return (averagePair,averageError)
+
+    # For the error case of a single row matrix being passed to the function
+        except IndexError:
+          coordPair = np.vstack(inputLocationTupleList[:, 0]).astype(np.float64)
+          errorArray = np.vstack(inputLocationTupleList[:, 1]).astype(np.float64)
+
+
+          return (coordPair,errorArray)
+ 


modules/geolocation/geolocation.py
```diff
@@ -129,3 +130,71 @@ def calculate_pixel_to_geo_mapping(self):
 
         # Return matrix product of mappedGeoMatrix and sourcePixelMatrixInverse
         return (mappedGeoMatrix.dot(sourcePixelMatrixInverse))
+    
+class OutputAnalysis:
```
Shrinjay, line 134: @Xierumeng Did you want this to be a seperate class? I think get_best_location should just be in the geolocation class, this kind of goes against my architecture

modules/geolocation/geolocation.py
```diff
@@ -129,3 +130,71 @@ def calculate_pixel_to_geo_mapping(self):
...
+            yCoordMedian = np.median(yCoord)
+            errorMedian = np.median(errorArray)
+
+    # A mmodified version of the trimmed mean calculation. Here the distance measuremnt is
+    # distance between every x,y point from the median coordinate pair, x median and y median.
+
+            for (x,y) in zip(xCoord,yCoord):
+                distance = math.sqrt((pow((x-xCoordMedian),2)+(pow((y-yCoordMedian),2))))
+
+                if distance<3.5:
```
Shrinjay, line 163: Is there any relevant literature specifying a distance of 3.5 as the cut-off for an outlier? Is it based off the 
```
x is an outlier IFF x > Q3 * 1.5 * InterQuartileRange
```
formula? 

From an engineering perspective, if we are setting a magic constant like this, put it in a constant called ```OUTLIER_DISTANCE = 3.5``` at the top of the python file

modules/geolocation/geolocation.py
```diff
@@ -129,3 +130,71 @@ def calculate_pixel_to_geo_mapping(self):
...
+
+
+            averagePair = (averageX,averageY)
+
+            return (averagePair,averageError)
+
+    # For the error case of a single row matrix being passed to the function
+        except IndexError:
+          coordPair = np.vstack(inputLocationTupleList[:, 0]).astype(np.float64)
+          errorArray = np.vstack(inputLocationTupleList[:, 1]).astype(np.float64)
```
Shrinjay, line 196: So in the error case we just don't do any processing?

modules/geolocation/geolocation.py
```diff
@@ -129,3 +130,71 @@ def calculate_pixel_to_geo_mapping(self):
...
+
+    def get_best_location(self,inputLocationTupleList):
+
+    # Splits the 3D numpy arrray into three separate arrays
+
+        try:
+            coordPair = np.vstack(inputLocationTupleList[:, 0]).astype(np.float64)
+            errorArray = np.vstack(inputLocationTupleList[:, 1]).astype(np.float64)
+            confidenceArray = np.vstack(inputLocationTupleList[:, 2]).astype(np.float64)
+
```
Shrinjay, line 144: Sorry if this is an uninformed question, but where are the values for errorArray and confidenceArray coming from? Can you give me an idea of what inputLocationTupleList looks like?

Shrinjay (commented):
Good work so far, just a few clarifications

modules/geolocation/geolocation.py
AddeshB: I had based the 3.5 off where I originally got the formula from, which recommended 3.5 as a threshold value for trimming using Z-scores, and what I had previously implemented was a modified version of Z-score.

modules/geolocation/geolocation.py
AddeshB: Here only a single row of data is being passed, so only one (x,y) pair, one confidence and one error. So in this case there's no calculation that can be done on only one value, so it gets returned

modules/geolocation/geolocation.py
AddeshB: An example of how the data would look= np.array(
[[10.4, 20.5], [2.4], [0.78]],
[[11.2, 19.9], [2.8], [0.69]],
…
[[10.1, 20.3], [1.9], [0.94]])

modules/geolocation/geolocation.py
AddeshB: I wasn't too sure either, initially I attempted to merge them together in one class but I got errors, so I had placed it in as a separate class. I can work on merging them together as a single class this week if they need to be separate

Xierumeng: Should be inside the `Geolocation` class. What's the problem?

modules/geolocation/geolocation.py
```diff
@@ -129,3 +130,71 @@ def calculate_pixel_to_geo_mapping(self):
...
         # Return matrix product of mappedGeoMatrix and sourcePixelMatrixInverse
         return (mappedGeoMatrix.dot(sourcePixelMatrixInverse))
+    
+class OutputAnalysis:
+
+    def get_best_location(self,inputLocationTupleList):
+
+    # Splits the 3D numpy arrray into three separate arrays
+
+        try:
```
Xierumeng, line 140: try-catch-exception blocks are *really* expensive, avoid them if at all possible. You can use an if statement to short-circuit the rest of the code.

modules/geolocation/geolocation.py
```diff
@@ -129,3 +130,71 @@ def calculate_pixel_to_geo_mapping(self):
...
+
+            xCoordMedian = np.median(xCoord)
+            yCoordMedian = np.median(yCoord)
+            errorMedian = np.median(errorArray)
+
+    # A mmodified version of the trimmed mean calculation. Here the distance measuremnt is
+    # distance between every x,y point from the median coordinate pair, x median and y median.
+
+            for (x,y) in zip(xCoord,yCoord):
+                distance = math.sqrt((pow((x-xCoordMedian),2)+(pow((y-yCoordMedian),2))))
```
Xierumeng, line 161: Consider using `math.hypot()`, from: https://stackoverflow.com/questions/5228383/how-do-i-find-the-distance-between-two-points

modules/geolocation/geolocation.py
```diff
@@ -129,3 +130,71 @@ def calculate_pixel_to_geo_mapping(self):
...
+
+                if distance<3.5:
+                    xCoordTrimmed.append(x)
+                    yCoordTrimmed.append(y)
+
+    # To handle the case where 2 points are too far apart from each other. Defaults to zero for both x and y
+
+                elif (distance>=3.5) and np.size(coordPair,0)==2:
```
Xierumeng, line 169: What if there are more than two coordinates but all of them are far away?

modules/geolocation/geolocation.py
```diff
@@ -129,3 +130,71 @@ def calculate_pixel_to_geo_mapping(self):
...
+
+
+    # Finds the average of the trimmed arrays using built-in np.average()
+
+            averageX = np.average(xCoordTrimmed, axis=0)
+            averageY = np.average(yCoordTrimmed, axis=0)
+            averageError = np.average(errorTrimmed, axis=0)
+
+
+            averagePair = (averageX,averageY)
```
Xierumeng, line 189: This creates a numpy array?

modules/geolocation/geolocation.py
Xierumeng: Are you using Z-scores right now? It doesn't look like it at the moment.

modules/geolocation/geolocation.py
Xierumeng: As I wrote above, move this to the beginning after an if statement to return early if there is only a single entry.

Note: You don't need to worry about being passed an empty array.

modules/geolocation/geolocation.py
```diff
@@ -129,3 +130,71 @@ def calculate_pixel_to_geo_mapping(self):
...
+    # To handle the case where 2 points are too far apart from each other. Defaults to zero for both x and y
+
+                elif (distance>=3.5) and np.size(coordPair,0)==2:
+                    print ("Warning, only 2 data points and they are too far apart")
+                    xCoordTrimmed.append(0)
+                    yCoordTrimmed.append(0)
+                    break
+
+            for e in errorArray:
+                distance = math.sqrt((pow((e-errorMedian),2)))
```
Xierumeng, line 176: Why are you calling functions when it's not necessary? Use the absolute function instead.

modules/geolocation/geolocation.py
```diff
@@ -129,3 +130,71 @@ def calculate_pixel_to_geo_mapping(self):
...
+
+    # To handle the case where 2 points are too far apart from each other. Defaults to zero for both x and y
+
+                elif (distance>=3.5) and np.size(coordPair,0)==2:
+                    print ("Warning, only 2 data points and they are too far apart")
+                    xCoordTrimmed.append(0)
+                    yCoordTrimmed.append(0)
+                    break
+
+            for e in errorArray:
```
Xierumeng, line 175: Instead of the for loop, try this: https://thispointer.com/delete-elements-from-a-numpy-array-by-value-or-conditions-in-python/

modules/geolocation/geolocation.py
AddeshB: That's an interesting case, in that case I'm considering defaulting it to (0,0) similar to what's been done here, except without the 2nd condition in the and statement

modules/geolocation/geolocation.py
Xierumeng: Use 7 then. There's going to be 3 clinics close together and we don't currently have a way of distinguishing them.

modules/geolocation/geolocation.py
Xierumeng: Default to the median instead.

modules/geolocation/geolocation.py
Xierumeng: Z-scores could be used if you saved all of the distances, but I'm sure you want to work on other tasks and we need to get other tasks done.

Add a `# TODO` comment stating this.

PR #35, opened 2021-03-26
Added test cases for Output Analysis

AddeshB:


modules/geolocation/unittests.py
```diff
@@ -202,6 +202,100 @@ def test_point_set_2(self):
         actual = self.locator.calculate_pixel_to_geo_mapping()
         # Test
         np.testing.assert_almost_equal(actual, expected)
+        
+        
+class TestOutputAnalysis(unittest.TestCase):
```
Shrinjay, line 207: I think we should standardize on pytest for testing

modules/geolocation/unittests.py
```diff
@@ -202,6 +202,100 @@ def test_point_set_2(self):
...
+    """
+    Tests Geolocation.outputanalysis(inputLocationTupleList)
+
+    """
+    def test_all_ones(self):
+        outputAnalysis = geolocation.OutputAnalysis()
+
+    # Setup
+        outputList = np.array([[[1, 1], [1], [1]]])
```
Shrinjay, line 216: To give any future developers a better idea of what's happening here, it would be better to define what expectedInputLocationTupleList first, and then define expectedCoordPair and expectedError from that

AddeshB: Got it, so a change in variable name from outputList to expectedInputLocationTupleList?

modules/geolocation/unittests.py
AddeshB: For the test cases I was following the format from the others that used unit test

Xierumeng, line 207: Name should be function/method that is being tested (especially since there isn't supposed to be a second class).

modules/geolocation/unittests.py
Xierumeng: No, maybe something like `listToAnalyze`, since it isn't an actual output from the function being tested.

modules/geolocation/unittests.py
Xierumeng: Out of scope of PR.

Am I hearing the dreaded word: *refactor*?

modules/geolocation/unittests.py
```diff
@@ -202,6 +202,100 @@ def test_point_set_2(self):
...
+
+        expectedCoordPair= (np.array([[1],  [1]]))
```
Xierumeng, line 218: Formatting nitpick: Spaces around the assignment operator.

Xierumeng: Also why is there a double space?

modules/geolocation/unittests.py
```diff
@@ -202,6 +202,100 @@ def test_point_set_2(self):
...
+        
+class TestOutputAnalysis(unittest.TestCase):
+    """
+    Tests Geolocation.outputanalysis(inputLocationTupleList)
+
+    """
+    def test_all_ones(self):
+        outputAnalysis = geolocation.OutputAnalysis()
+
+    # Setup
```
Xierumeng, line 215: Formatting nitpick: Comments should be at the same indentation level as the code around it.

modules/geolocation/unittests.py
```diff
@@ -202,6 +202,100 @@ def test_point_set_2(self):
...
+        outputList = np.array([[[1, 1], [1], [1]]])
+
+        expectedCoordPair= (np.array([[1],  [1]]))
+        expectedError= (np.array([1]))
+
+        # Run
+        actual = outputAnalysis.get_best_location(outputList)
+
+        np.testing.assert_equal(actual[0], expectedCoordPair)
```
Xierumeng, line 224: Because we are working with floating point numbers, use `assert_almost_equal()` instead.

PR #36, opened 2021-03-27
Collinear points

manumanuk:
Purpose:
Created get_non_collinear_points() function to geolocation module, which grabs a group of 4 points which are non-collinear

New Changes:
- Completed get_non_collinear_points() function
- Added documentation to get_non_collinear_points()
- Added unit tests

Outstanding Changes:
- Clarification needed on whether a return value is supposed to be used for the output, or whether one of the member variables is to be altered

### DIFF
--- a/modules/geolocation/geolocation.py
+++ b/modules/geolocation/geolocation.py
@@ -117,9 +117,11 @@ class Geolocation:
         """
         x = 0
         y = 1
-        return 0.5*(p1[x]*(p2[y]-p3[y]) +
-                    p2[x]*(p3[y]-p1[y]) +
-                    p3[x]*(p1[y]-p2[y])) == 0
+        # Calculates the area of a triangle and checks if this value is 0
+        # Actually calculates 2 times the area, since it skips the unnecessary step of multiplication by 0.5
+        return (p1[x]*(p2[y]-p3[y]) +
+                p2[x]*(p3[y]-p1[y]) +
+                p3[x]*(p1[y]-p2[y])) == 0
 
 
     def get_non_collinear_points(self, coordinatesArray):
@@ -137,33 +139,35 @@ class Geolocation:
             Array with dimensions (4, 2), containing a list of coordinates that are non-collinear,
             or an empty list if none were found in the input array
         """
+        NUM_POINTS_NEEDED = 4
         # Empty array in case no set of four non-collinear points are found
-        collinearPoints = np.zeros(shape=(4, 2))
+        collinearPoints = np.empty(shape=(0, 2))
 
         # If there aren't four points, return the empty array
-        if len(coordinatesArray) < 4:
+        if len(coordinatesArray) < NUM_POINTS_NEEDED:
             return collinearPoints
 
         # Look at all sequential pairs of four points
         for i in range(0, len(coordinatesArray)):
             # Array for storing the four points currently being considered
-            points = np.zeros(shape=(4, 2))
+            points = np.empty(shape=(NUM_POINTS_NEEDED, 2))
 
             # Append four sequential points to the array, loop around to index 0 if needed
-            for j in range(0, 4):
+            # For efficiency, this algorithm will check through sequential sets of points only, rather than
+            # testing every single combination.
+            for j in range(0, NUM_POINTS_NEEDED):
                 points[j] = coordinatesArray[(i+j) % len(coordinatesArray)]
 
             # Check collinearity of all possible combinations
-            areFourCollinear = True
-            for i in range(0, 4):
-                areFourCollinear &= self.__are_three_points_collinear(points[i], 
-                                                                      points[(i+1) % 4],
-                                                                      points[(i+2) % 4])
+            areNotFourCollinear = True
+            for i in range(0, NUM_POINTS_NEEDED):
+                areNotFourCollinear &= not self.__are_three_points_collinear(points[i], 
+                                                                             points[(i+1) % NUM_POINTS_NEEDED],
+                                                                             points[(i+2) % NUM_POINTS_NEEDED])
             
             # If all four points are non-collinear, return this combination of points
-            if not areFourCollinear:
-                collinearPoints = points
-                break
+            if areNotFourCollinear:
+                return points
 
         return collinearPoints
 
--- a/modules/geolocation/geolocation.py
+++ b/modules/geolocation/geolocation.py
@@ -2,7 +2,6 @@
 Geolocation module to map pixel coordinates to geographical coordinates
 """
 
-import itertools
 import numpy as np
 
 class Geolocation:
@@ -112,20 +111,22 @@ class Geolocation:
             Array with dimensions (4, 2), containing a list of coordinates that are non-collinear,
             or an empty list if none were found in the input array
         """
-        collinearPoints = np.empty(shape=(4, 2))
-        combos = itertools.combinations(coordinatesArray, 4)
+        collinearPoints = np.zeros(shape=(4, 2))
 
-        for combo in combos:
-            x1, y1 = combo[0]
-            x2, y2 = combo[1]
-            x3, y3 = combo[2]
-            x4, y4 = combo[3]
+        if len(coordinatesArray) < 4:
+            return collinearPoints
+
+        for i in range(0, len(coordinatesArray)-3):
+            x1, y1 = coordinatesArray[i]
+            x2, y2 = coordinatesArray[i+1]
+            x3, y3 = coordinatesArray[i+2]
+            x4, y4 = coordinatesArray[i+3]
 
             if (0.5*(x1*(y2-y3)+x2*(y3-y1)+x3*(y1-y2)) != 0 and
                 0.5*(x2*(y3-y4)+x3*(y4-y2)+x4*(y2-y3)) != 0 and
                 0.5*(x3*(y4-y1)+x4*(y1-y3)+x1*(y3-y4)) != 0 and
                 0.5*(x4*(y1-y2)+x1*(y2-y4)+x2*(y4-y1)) != 0):
-                print("True")
+                
                 collinearPoints = np.array([[x1, y1], [x2, y2], [x3, y3], [x4, y4]])
                 break
 
--- a/modules/geolocation/geolocation.py
+++ b/modules/geolocation/geolocation.py
@@ -96,6 +96,14 @@ class Geolocation:
 
         return pixelGeoPairs
     
+    def __are_three_points_collinear(self, p1, p2, p3):
+        x = 0
+        y = 1
+        return 0.5*(p1[x]*(p2[y]-p3[y]) +
+                    p2[x]*(p3[y]-p1[y]) +
+                    p3[x]*(p1[y]-p2[y])) == 0
+
+
     def get_non_collinear_points(self, coordinatesArray):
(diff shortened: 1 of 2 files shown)


modules/geolocation/geolocation.py
```diff
@@ -96,6 +95,42 @@ def gather_point_pairs(self):
...
+
+        if len(coordinatesArray) < 4:
+            return collinearPoints
+
+        for i in range(0, len(coordinatesArray)-3):
+            x1, y1 = coordinatesArray[i]
+            x2, y2 = coordinatesArray[i+1]
+            x3, y3 = coordinatesArray[i+2]
+            x4, y4 = coordinatesArray[i+3]
+
```
Shrinjay, line 124: I think the readability and brevity of this code could be improved. For starters, make x and y numpy arrays. Then structure lines 120-123 as:
`
for j in range(0, 4):
x[j], y[j] = coordinatesArray[i + j]
`
which should result in the same behaviour as:
x[0], y[0] = coordinatesArray[i]
...
x[3], y[3] = coordinatesArray[i+3]

modules/geolocation/geolocation.py
```diff
@@ -96,6 +95,42 @@ def gather_point_pairs(self):
...
+            if (0.5*(x1*(y2-y3)+x2*(y3-y1)+x3*(y1-y2)) != 0 and
+                0.5*(x2*(y3-y4)+x3*(y4-y2)+x4*(y2-y3)) != 0 and
+                0.5*(x3*(y4-y1)+x4*(y1-y3)+x1*(y3-y4)) != 0 and
+                0.5*(x4*(y1-y2)+x1*(y2-y4)+x2*(y4-y1)) != 0):
+
```
Shrinjay, line 129: Likewise, lines 125-129 should be pulled out into their own functions, and implemented like:

```
def checkColinear(x, y):

all_colinear = False

for i in range(0, 4): 

  i_mod = i % 5
  i_1_mod = (i+1) % 5
  i_2_mod = (i+2) % 5
  all_colinear = all_colinear AND ( 0.5 * (x[i_mod]*(y[i_1_mod] - y[i_2_mod])+x[i_1_mod]*(y[i_2_mod] - y[i_mod]) + x[i_2_mod]*(y[i_mod] - y[i_1_mod]) != 0)

return all_colinear
```
Feel free to change up or challenge the implementation in both my recommendations

modules/geolocation/geolocation.py
```diff
@@ -96,6 +95,42 @@ def gather_point_pairs(self):
...
+            x2, y2 = coordinatesArray[i+1]
+            x3, y3 = coordinatesArray[i+2]
+            x4, y4 = coordinatesArray[i+3]
+
+            if (0.5*(x1*(y2-y3)+x2*(y3-y1)+x3*(y1-y2)) != 0 and
+                0.5*(x2*(y3-y4)+x3*(y4-y2)+x4*(y2-y3)) != 0 and
+                0.5*(x3*(y4-y1)+x4*(y1-y3)+x1*(y3-y4)) != 0 and
+                0.5*(x4*(y1-y2)+x1*(y2-y4)+x2*(y4-y1)) != 0):
+                
+                collinearPoints = np.array([[x1, y1], [x2, y2], [x3, y3], [x4, y4]])
```
Shrinjay, line 130: Suggestion here, thought not a big deal:

```
collinearPoints = np.array([[x[i], y[i]] for i in range(0, 4)])
```
Also, why an array instead of a tuple for storing the points?

Shrinjay (commented):
Good stuff, just some thoughts

modules/geolocation/geolocation.py
manumanuk: I think this does make sense for readability; do you think I should define the function within get_non_collinear_points() or as a new class member?

manumanuk:
PR has been edited:
- New implementation for better code readability
- All unit tests still pass

modules/geolocation/geolocation.py
```diff
@@ -96,6 +95,77 @@ def gather_point_pairs(self):
...
+            Array with dimensions ( , 2), containing a list of coordinates
+
+        Returns
+        -------
+        np.ndarray
+            Array with dimensions (4, 2), containing a list of coordinates that are non-collinear,
+            or an empty list if none were found in the input array
+        """
+        # Empty array in case no set of four non-collinear points are found
+        collinearPoints = np.zeros(shape=(4, 2))
```
Xierumeng, line 141: Return np.empty(shape=(0, 2)) on failure (array of size 0).

modules/geolocation/geolocation.py
```diff
@@ -96,6 +95,77 @@ def gather_point_pairs(self):
...
+
+            # Check collinearity of all possible combinations
+            areFourCollinear = True
+            for i in range(0, 4):
+                areFourCollinear &= self.__are_three_points_collinear(points[i], 
+                                                                      points[(i+1) % 4],
+                                                                      points[(i+2) % 4])
+            
+            # If all four points are non-collinear, return this combination of points
+            if not areFourCollinear:
```
Xierumeng, line 164: Can short-circuit by returning directly here instead of breaking out of the loop.

modules/geolocation/geolocation.py
```diff
@@ -96,6 +95,77 @@ def gather_point_pairs(self):
...
+        """
+        # Empty array in case no set of four non-collinear points are found
+        collinearPoints = np.zeros(shape=(4, 2))
+
+        # If there aren't four points, return the empty array
+        if len(coordinatesArray) < 4:
+            return collinearPoints
+
+        # Look at all sequential pairs of four points
+        for i in range(0, len(coordinatesArray)):
```
Xierumeng, line 148: I can make this algorithm fail, but the circumstances in which it does is even less likely than my original one. It is slightly slower (not really an issue given how few points it will be called with, but just a consideration).

Add a comment explaining why sequential (rather than all combinations).

(Not part of PR) Note: I believe an actual correct algorithm must at minimum check all combinations, which means the problem is NP (solvable in exponential-time, verifiable in polynomial-time).

Xierumeng: That is to say, add the comment, no need to change anything.

modules/geolocation/geolocation.py
```diff
@@ -96,6 +95,77 @@ def gather_point_pairs(self):
...
+            # Array for storing the four points currently being considered
+            points = np.zeros(shape=(4, 2))
```
Xierumeng, line 150: Micro-optimization: np.empty() to skip initialization.

modules/geolocation/geolocation.py
```diff
@@ -96,6 +95,77 @@ def gather_point_pairs(self):
...
+        -------
+        np.ndarray
+            Array with dimensions (4, 2), containing a list of coordinates that are non-collinear,
+            or an empty list if none were found in the input array
+        """
+        # Empty array in case no set of four non-collinear points are found
+        collinearPoints = np.zeros(shape=(4, 2))
+
+        # If there aren't four points, return the empty array
+        if len(coordinatesArray) < 4:
```
Xierumeng, line 144: I see a lot of "4" as a magic number, consider making it a variable/constant.

modules/geolocation/geolocation.py
```diff
@@ -128,4 +198,4 @@ def calculate_pixel_to_geo_mapping(self):
         sourcePixelMatrixInverse = np.linalg.pinv(sourcePixelMatrix)
 
         # Return matrix product of mappedGeoMatrix and sourcePixelMatrixInverse
-        return (mappedGeoMatrix.dot(sourcePixelMatrixInverse))
+        return (mappedGeoMatrix.dot(sourcePixelMatrixInverse))
```
Xierumeng, line 201: Formatting: Add an empty line at the end of the file.

modules/geolocation/unittests.py
```diff
@@ -121,6 +121,89 @@ def test_camera_offset_from_origin_pointing_sideways_with_not_enough_downward_pi
...
+        coordinatesArray = np.array([[0, 0], [1, 1], [-1, -1]])
+        expected = np.empty(shape=(4, 2))
+
+        # Run
+        actual = self.locator.get_non_collinear_points(coordinatesArray)
+
+        # Test
+        np.testing.assert_almost_equal(actual, expected)
+
+    def less_than_four_points_non_collinear(self):
```
Xierumeng, line 154: Duplicate test case.

modules/geolocation/unittests.py
```diff
@@ -121,6 +121,89 @@ def test_camera_offset_from_origin_pointing_sideways_with_not_enough_downward_pi
...
+
+        # Run
+        actual = self.locator.get_non_collinear_points(coordinatesArray)
+
+        # Test
+        np.testing.assert_almost_equal(actual, expected)
+
+    
+    def test_more_than_four_points_no_correct_cases(self):
+        coordinatesArray = np.array([[0, 0], [1, 1], [2, 2], [10, 10], [15, 15], [0, 0]])
```
Xierumeng, line 197: Add one point outside of this line of points to make things slightly more interesting.

Xierumeng: Also, there is a duplicate [0, 0], why is that?

manumanuk:
Fixed all issues discussed above

modules/geolocation/geolocation.py
```diff
@@ -96,6 +95,81 @@ def gather_point_pairs(self):
...
+            Array with dimensions (4, 2), containing a list of coordinates that are non-collinear,
+            or an empty list if none were found in the input array
+        """
+        NUM_POINTS_NEEDED = 4
+        # Empty array in case no set of four non-collinear points are found
+        collinearPoints = np.empty(shape=(0, 2))
+
+        # If there aren't four points, return the empty array
+        if len(coordinatesArray) < NUM_POINTS_NEEDED:
+            return collinearPoints
```
Xierumeng, line 148: collinearPoints is just an array of size 0, why not `return np.empty(shape=(0, 2))` directly?

modules/geolocation/geolocation.py
```diff
@@ -96,6 +95,81 @@ def gather_point_pairs(self):
...
+            for i in range(0, NUM_POINTS_NEEDED):
+                areNotFourCollinear &= not self.__are_three_points_collinear(points[i], 
+                                                                             points[(i+1) % NUM_POINTS_NEEDED],
+                                                                             points[(i+2) % NUM_POINTS_NEEDED])
+            
+            # If all four points are non-collinear, return this combination of points
+            if areNotFourCollinear:
+                return points
+
+        return collinearPoints
```
Xierumeng, line 172: Again, could return directly since `points` is the other case.

modules/geolocation/unittests.py
```diff
@@ -121,6 +121,81 @@ def test_camera_offset_from_origin_pointing_sideways_with_not_enough_downward_pi
...
+    Tests get_non_collinear_points()
+    """
+    def setUp(self):
+        self.locator = geolocation.Geolocation()
+        return
+    
+    def test_empty_input_array(self):
+        # Setup
+        coordinatesArray = np.array([])
+        expected = 0
```
Xierumeng, line 135: Make this consistent with the failures below.

Xierumeng:
@Shrinjay Approved.

PR #37, opened 2021-03-30
Input conversion

andrew2002zhao:
-takes geolocation data and returns a set of vectors cooresponding to the plane
-returns an o vector, the vector that points to the plane in world space
-returns c, u and v, vectors that constitute the direction vectors for the camera in world space.
-used 4 unit tests to affirm that the methods all work


modules/geolocation/geolocation.py
```diff
@@ -129,3 +140,210 @@ def calculate_pixel_to_geo_mapping(self):
...
+        list:
+            list of numpy arrays that represent the o, c, u and v vectors
+        """
+
+        # get plane and camera rotation matrices
+        planeRotation = self.__calculate_rotation_matrix(eulerPlane)
+        cameraRotation = self.__calculate_rotation_matrix(eulerCamera)
+
+        # get plane and camera compound rotation matrix
+        # note: assume apply plane -> camera rotations in that order
```
Xierumeng, line 180: Isn't it the other way around? The code is also different from this comment.

modules/geolocation/geolocation.py
```diff
@@ -129,3 +140,210 @@ def calculate_pixel_to_geo_mapping(self):
...
+        compoundRotationMatrix = planeRotation.dot(cameraRotation)
```
Xierumeng, line 181: Use `@` or `numpy.matmul()`.

modules/geolocation/geolocation.py
```diff
@@ -129,3 +140,210 @@ def calculate_pixel_to_geo_mapping(self):
...
+
+        # calculate gps module to camera offset
+        gpsCameraOffset = np.add(gpsOffset, cameraOffset)
```
Xierumeng, line 184: Shouldn't this be a subtraction?

modules/geolocation/geolocation.py
```diff
@@ -129,3 +140,210 @@ def calculate_pixel_to_geo_mapping(self):
...
+        origin:numpy array
+            Array containing the origin of the world space
+
+        Returns
+        -------
+        oVector: numpy array
+            Array containing components of the o vector
+        """
+        # get GPS module coordinates
+        gpsModule = np.zeros(3)
```
Xierumeng, line 221: Micro-optimization: `numpy.empty()` so we're not initializing when unneeded.

modules/geolocation/geolocation.py
```diff
@@ -129,3 +140,210 @@ def calculate_pixel_to_geo_mapping(self):
...
         return (mappedGeoMatrix.dot(sourcePixelMatrixInverse))
+
+
+    def convert_input(self, latitude: int, longitude: int, altitude: int, worldOrigin: np.ndarray, gpsOffset: np.ndarray, cameraOffset: np.ndarray, eulerCamera: dict, eulerPlane: dict):
+        """
+        Converts telemtry data into workable data
+
+        Parameters
+        ----------
+        latitude: int
```
Xierumeng, line 151: These may be floating points, not ints.

modules/geolocation/geolocation.py
```diff
@@ -129,3 +140,210 @@ def calculate_pixel_to_geo_mapping(self):
...
+        gpsModule[2] = origin[2] + altitude
+
+        # transform gps-to-camera offset from plane space to world space by applying inverse
+        # rotation matrix using (AB)^-1 = B^-1 A^-1
+        rotationMatrix = np.matmul(planeRotation, cameraRotation)
+        transposedMatrix = np.transpose(rotationMatrix)
+        rotatedOffset = transposedMatrix.dot(gpsCameraOffset)
+
+        # get camera coordinates in world space
+        oVector = np.add(rotatedOffset, gpsModule)
```
Xierumeng, line 233: This could be returned directly.

modules/geolocation/geolocation.py
```diff
@@ -129,3 +140,210 @@ def calculate_pixel_to_geo_mapping(self):
...
+        cVector:
+            Array containing components of the c vector
+        """
+        # apply rotation matrix to camera direction
+        # note: this assumes that camera euler angles are w.r.t. plane space
+        cVector = compoundRotationMatrix.dot(cVectorCameraSpace)
+
+        # normalize since magnitude doesn't matter
+        norm = np.linalg.norm(cVector)
+        if(norm != 0) :
```
Xierumeng, line 259: Formatting nitpick: Space between if and open parantheses, no space between closing parantheses and colon.

modules/geolocation/geolocation.py
```diff
@@ -129,3 +140,210 @@ def calculate_pixel_to_geo_mapping(self):
...
+        # note: this assumes that camera euler angles are w.r.t. plane space
+        uVector = compoundRotationMatrix.dot(uVectorCameraSpace)
+
+        # normalize output since camera direction magnitude doesn't matter
+        norm = np.linalg.norm(uVector)
+        if (norm != 0) :
+            uVector = uVector / norm
+
+        uVector = np.squeeze(uVector)
+        return uVector
```
Xierumeng, line 291: This should be multiplied by a field-of-view constant so that the ratio between this and the **c** vector are correct.

modules/geolocation/geolocation.py
```diff
@@ -129,3 +140,210 @@ def calculate_pixel_to_geo_mapping(self):
...
+        u: numpy array
+            Array containing one camera rotation vector
+
+        Returns
+        -------
+        numpyArray:
+            array containing the remaining camera rotation vector
+        """
+        # cross product c and u to get v
+        return np.cross(c, u)
```
Xierumeng, line 310: Also FOV factor needed here.

modules/geolocation/unittests.py
```diff
@@ -195,7 +195,131 @@ def test_point_set_2(self):
...
+        longitude = 0
+        altitude = 0
+        worldOrigin = np.array([0, 0, 0])
+        gpsOffset = np.array([0, 0, 0])
+        cameraOffset = np.array([0, 0, 0])
+        eulerCamera = {"x": 0, "y": 0, "z": 0}
+        eulerPlane = {"x": 0, "y": 0, "z": 0}
+
+
+        geoLocationClass = geolocation.Geolocation()
```
Xierumeng, line 216: This can be placed in the `setUp()` method (see e.g. https://github.com/UWARG/computer-vision-python/pull/26/files#diff-d96e1349803cbb8d531f917f69838c18e71873ba712c19f8977f52a4a8702a30R202).

modules/geolocation/geolocation.py
```diff
@@ -129,3 +140,210 @@ def calculate_pixel_to_geo_mapping(self):
...
+        """
+        # get euler angles
+        # note: naming conventions specified in CV-Telemetry docs and commandModule specs
+        yawAngle = eulerAngles["z"]
+        pitchAngle = eulerAngles["y"]
+        rollAngle = eulerAngles["x"]
+
+        # get plane yaw rotation matrix
+        rollRotation = np.array([[1, 0, 0],
+                                [0, float(np.cos(np.deg2rad([rollAngle]))), -1 * float(np.sin(np.deg2rad([rollAngle])))],
```
Xierumeng, line 334: There's a lot of converting to radians, why not do it when the variable is first declared and initialized?

modules/geolocation/geolocation.py
```diff
@@ -129,3 +140,210 @@ def calculate_pixel_to_geo_mapping(self):
...
+        rotationMatrix: numpy array
+            Rotation matrix calculated from the given euler angles
+        """
+        # get euler angles
+        # note: naming conventions specified in CV-Telemetry docs and commandModule specs
+        yawAngle = eulerAngles["z"]
+        pitchAngle = eulerAngles["y"]
+        rollAngle = eulerAngles["x"]
+
+        # get plane yaw rotation matrix
```
Xierumeng, line 332: Why are the comments different from the code?

modules/geolocation/geolocation.py
```diff
@@ -129,3 +140,210 @@ def calculate_pixel_to_geo_mapping(self):
...
+        rollRotation = np.array([[1, 0, 0],
```
Xierumeng, line 333: Reorder yaw and pitch rotation matrices so they're in order of the matrix multiplication.

modules/geolocation/geolocation.py
```diff
@@ -129,3 +140,210 @@ def calculate_pixel_to_geo_mapping(self):
...
+        # get plane roll rotation matrix
+        yawRotation = np.array([[float(np.cos(np.deg2rad([yawAngle]))), -1 * float(np.sin(np.deg2rad([yawAngle]))), 0],
+                                 [float(np.sin(np.deg2rad([yawAngle]))), float(np.cos(np.deg2rad([yawAngle]))), 0],
+                                 [0, 0, 1]])
+
+        # get total plane rotation matrix based on euler rotation theorem
+        # note: assume ZYX euler angles (i.e. R = RxRyRz); otherwise, change the multiplication order
+        rotationMatrix = np.matmul(np.matmul(yawRotation, pitchRotation), rollRotation)
+
+        return rotationMatrix
```
Xierumeng, line 349: Formatting nitpick: Files should end with an empty line.

modules/geolocation/geolocation.py
ksisjaya: i thought it should be plane rotation first then camera rotation since the camera euler angles are wrt plane space. so you'd need to transform into plane space first before applying the camera rotation?

either way, yes, the comment doesn't match the code

modules/geolocation/geolocation.py
ksisjaya: CV-Telemetry docs show euler angles from firmware will be in radians. will remove deg2rad conversion instead

modules/geolocation/geolocation.py
ksisjaya: yes, it should be

modules/geolocation/geolocation.py
```diff
@@ -129,3 +140,210 @@ def calculate_pixel_to_geo_mapping(self):
...
+            Array containing the origin of the world space
+
+        Returns
+        -------
+        oVector: numpy array
+            Array containing components of the o vector
+        """
+        # get GPS module coordinates
+        gpsModule = np.zeros(3)
+        gpsModule[0] = origin[0] + latitude
```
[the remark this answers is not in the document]
ksisjaya: yes, will change

modules/geolocation/geolocation.py
```diff
@@ -129,3 +143,178 @@ def calculate_pixel_to_geo_mapping(self):
...
+        uVector = np.matmul(compoundRotationMatrix, self.__U_VECTOR_CAMERA_SPACE)
+
+        # normalize output since camera direction magnitude doesn't matter
+        norm = np.linalg.norm(uVector)
+        if (norm != 0):
+            uVector = uVector / norm
+
+        # reshape vector and apply field of view factor
+        uVector = np.squeeze(uVector)
+        return uVector * self.__FOV_FACTOR
```
Xierumeng, line 257: So there's both a horizontal FOV and vertical FOV, sorry. We also need the magnitude of **c** (unless ||**c**|| is always 1, in which case write a comment).

FOV_FACTOR_H would be the constant ratio ||**u**|| / ||**c**||, FOV_FACTOR_V would be the constant ratio ||**v**|| / ||**c**||.

modules/geolocation/geolocation.py
```diff
@@ -129,3 +143,178 @@ def calculate_pixel_to_geo_mapping(self):
...
+        Returns
+        -------
+        numpyArray:
+            array containing the remaining camera rotation vector
+        """
+        # cross product c and u to get v
+        v_vector = np.cross(c, u)
+        
+        # apply field of view factor
+        return v_vector * self.__FOV_FACTOR
```
Xierumeng, line 279: Same as above comment.

modules/geolocation/geolocation.py
ksisjaya: **c** and **v** are normalized already. will change the comment to make it clearer that their magnitudes = 1. likewise, will also add normalization to **u**.

this makes it easier to calculate the final uVector & vVector. just so we're all on the same page, below is an example calc: (please lmk if any of this is wrong @Xierumeng)

let ||**Ufinal**|| / ||**C**|| = FOV_FACTOR_H -> (FOV_FACTOR condition you mentioned)
let **Ufinal** = **Uinitial** * x, x = scalar multiplier to be found

∴ ||**Uinitial** * x|| / ||**C**|| = FOV_FACTOR_H
x * ||**Uinitial**|| / ||**C**|| = FOV_FACTOR_H
Given **C** and **Uinitial** normalized, ||**C**|| = 1, ||**Uinitial**|| = 1
∴ x = FOV_FACTOR_H
∴ **Ufinal** = **Uinitial** * FOV_FACTOR_H

so resulting logic is:
`uVector = normalize(uVector) `
`return uVector * FOV_FACTOR_H`
same calcs apply to vVector

Xierumeng: Looks good.

modules/geolocation/unittests.py
```diff
@@ -195,7 +199,104 @@ def test_point_set_2(self):
...
         
- 
+    def test_90_degree_rotation(self):
+
+        expected_o = np.zeros(3)
+        expected_c = np.array([-1, 0, 0])
+        expected_u = np.array([0, 1, 0])
+        expected_v = np.array([0, 0, -1])
+
+        self.locator._Geolocation__FOV_FACTOR = 1
```
Xierumeng, line 237: Update unit tests.

modules/geolocation/geolocation.py
Xierumeng: Update this comment.

modules/geolocation/geolocation.py
```diff
@@ -129,3 +144,184 @@ def calculate_pixel_to_geo_mapping(self):
...
+        uVector = np.matmul(compoundRotationMatrix, self.__U_VECTOR_CAMERA_SPACE)
+
+        # normalize uVector so ||uVector|| = 1
+        norm = np.linalg.norm(uVector)
+        if (norm != 0):
+            uVector = uVector / norm
+
+        # reshape vector and apply field of view factor
+        uVector = np.squeeze(uVector)
+        return uVector * self.__FOV_FACTOR_H
```
Xierumeng, line 258: Add a comment explaining why we don't need **c** here (i.e. because magnitude of **c** is 1).

modules/geolocation/unittests.py
Xierumeng: Also, use non-identity (non-1) values (so it's tested).

Xierumeng:
@Shrinjay Approved.

PR #39, opened 2021-04-02
Added output analysis as a class method

AddeshB:
Also fixed and added certain parts based off comments from the previous PR, notably:
1. Added get_best_location as a class method for geolocation class
2. Used functions such as math.hypot() to compute distance between points, so math is imported
3. Added an inner helper function to handle the error case where all values are too far apart resulting in the trimmed array being empty, and defaults to the median value as the final result. Otherwise it calculates the average of the trimmed array. 
4. Removed try-catch-exception block and replaced with if-else statement 
5. Formatting errors fixed

### DIFF
--- a/modules/geolocation/geolocation.py
+++ b/modules/geolocation/geolocation.py
@@ -3,6 +3,7 @@ Geolocation module to map pixel coordinates to geographical coordinates
 """
 
 import numpy as np
+import math
 
 
 class Geolocation:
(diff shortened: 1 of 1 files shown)


modules/geolocation/geolocation.py
```diff
@@ -129,3 +130,61 @@ def calculate_pixel_to_geo_mapping(self):
 
         # Return matrix product of mappedGeoMatrix and sourcePixelMatrixInverse
         return (mappedGeoMatrix.dot(sourcePixelMatrixInverse))
+
+    def get_best_location(self,inputLocationTupleList):
+
+        # Inner helper function that checks if the trimmed array is empty ie all values were too far apart from the median
+        # Returns median if array is empty, otherwise finds the average of the trimmed array
+        def check_for_empty(trimmedArray,median):
```
Xierumeng, line 138: Move this outside of the get_best_location() function, add two underscores: `__check_for_empty()` to indicate that it is private.

Also, a better name for the function may be: `__get_average_otherwise_median()`

modules/geolocation/geolocation.py
```diff
@@ -129,3 +130,61 @@ def calculate_pixel_to_geo_mapping(self):
...
+
+        # For the  case of a single row matrix being passed to the function
+        if np.size(inputLocationTupleList,0)==1:
+
+            coordPair = np.vstack(inputLocationTupleList[:, 0]).astype(np.float64)
+            errorArray = np.vstack(inputLocationTupleList[:, 1]).astype(np.float64)
+
+            return (np.array(coordPair),np.array(errorArray))
+
+        else:
```
Xierumeng, line 153: Not required because the return short-circuits the entire function.

PR #40, opened 2021-04-02
Updated test cases for output analysis

AddeshB:
Aside from formatting errors, test cases were added to handle exception cases where there are one or more outliers in error array

### DIFF
--- a/modules/geolocation/unittests.py
+++ b/modules/geolocation/unittests.py
@@ -203,6 +203,134 @@ class TestPointMatrixToGeoMapping(unittest.TestCase):
         # Test
         np.testing.assert_almost_equal(actual, expected)
 
+class TestGetBestLocation(unittest.TestCase):
+    """
+    Tests Geolocation.outputanalysis(inputLocationTupleList)
+    """
+    def test_all_ones(self):
+
+        # Setup
+        outputAnalysis = geolocation.Geolocation()
+
+        listToAnalyze = np.array([[[1, 1], [1], [1]]])
+
+        expectedCoordPair = np.array([[1,1]])
+        expectedError = np.array([[1]])
+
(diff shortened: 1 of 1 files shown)


modules/geolocation/unittests.py
```diff
@@ -203,6 +203,134 @@ def test_point_set_2(self):
...
+        np.testing.assert_almost_equal(actual[0], expectedCoordPair)
+        np.testing.assert_almost_equal(actual[1], expectedError)
+
+    def test_all_inputs_too_far_from_each_other(self):
+
+        # Setup
+        outputAnalysis = geolocation.Geolocation()
+        listToAnalyze = np.array([
+        [[10, 20], [2.4], [0.78]],
+        [[100, 200], [1.9], [0.94]]])
```
Xierumeng, line 285: Add at least one extra element as currently the average and the median are the same (so this test cannot tell if it's actually getting the median).

modules/geolocation/unittests.py
```diff
@@ -203,6 +203,134 @@ def test_point_set_2(self):
...
+        expectedCoordPair = (np.array([[11],[20]]))
+        expectedError = (np.array([2.2]))
+
+        # Run
+        actual = outputAnalysis.get_best_location(listToAnalyze)
+
+        np.testing.assert_almost_equal(actual[0], expectedCoordPair)
+        np.testing.assert_almost_equal(actual[1], expectedError)
+
+    def test_input_with_one_outlier(self):
```
Xierumeng, line 261: Rather than "input", you could use "point" or "coordinate" (e.g. `test_point_with_one_outlier()` or `test_coordinate_with_one_outlier()`), because at the moment you're not testing the errors, which I saw was further below.

PR #41, opened 2021-04-03
Added output analysis as a class method

AddeshB:
New changes:
1. Removed inner helper function and added it as a private class method
2. Removed return statement in the 'if' block
3. Name changes


Xierumeng:
@Shrinjay Approved.

PR #43, opened 2021-04-04
Prevent exception handling from terminating execution

ksisjaya:
- Change exception handling to logger instead
- **_Return NoneType from command module to indicate an error has occurred_**
- Change to modular logger
- Update test cases to check for logger message
- Update test and debug log
- Allow for logging in test cases to log to debug.log file

### DIFF
--- a/modules/commandModule/commandModule.py
+++ b/modules/commandModule/commandModule.py
@@ -73,7 +73,7 @@ class CommandModule:
         pigoFileDirectory: str
             String of POGI file directory
         """
-        self.__logger = logging.getLogger()
+        self.__logger = logging.getLogger(__name__)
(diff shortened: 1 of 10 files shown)


ksisjaya:
@Shrinjay here's changes to prevent commandModule from terminating execution on raising an exception.

note:
- execution will be terminated if json filepaths are not set properly; if not, then data will not be read/written at all (if you also want to suppress these exceptions please lmk)

Shrinjay (approved):
I was thinking more on having error handling rather than just returning None, but because of our timeline I'll let this pass for now. I'll add some interface code between this class and the pipelines so None is not let into the pipeline to compensate. Thanks for taking care of this.

PR #45, opened 2021-04-14
Added values for distance calculation

thrvg:
Added some values required for distance calculation. Previously, I had put 1 as a placeholder there.


modules/targetAcquisition/taxi/Taxi.py
```diff
@@ -95,16 +95,16 @@ def calculate_distance(pts):
...
         # (Known) real Height of the box in mm
         realHeight = 101.6
 
         # (Known) height of the image in pixels
-        imageHeight = 1
+        imageHeight = 1080
 
         # (Known) height of the sensor in mm
-        sensorHeight = 1
+        sensorHeight = 4.63
```
Shrinjay, line 107: Nitpick here, but would be nice if you could move these values into constant variables declared in CONSTANT_CASE (example of constant case there) at the top of the file.

Shrinjay (changes requested):

modules/targetAcquisition/taxi/Taxi.py
thrvg: Changed to CONSTANT_CASE!

thrvg:
Hey I put the CONSTANTS at the top

Shrinjay (approved):
Sorry for the late review, LGTM now

PR #46, opened 2021-04-26
Geolocation workers

Xierumeng:


modules/geolocation/geolocation.py
```diff
@@ -411,6 +413,17 @@ def __get_average_otherwise_median(self,trimmedArray,median):
             return np.average(trimmedArray,axis=0)
 
 
+    # Helper function required for input
+    def concatenate_locations(self, newLocations):
+
+        self.__locationsList = self.__locationsList + newLocations
+
+
+    def convert_NP_from_list(self, locationsList):
```
Shrinjay, line 422: Does this really need to be a wrapper function rather than just a direct call to np.array(...)? Only seems to be used once right now.

modules/geolocation/geolocationWorker.py
```diff
@@ -0,0 +1,89 @@
...
+import multiprocessing as mp
+# For exception handling
+import queue
+
+
+def exit_requested(requestQueue):
+    return not requestQueue.empty()
+
+
+def geolocation_worker(pause, exitRequest, pipelineIn, pipelineOut, pipelineOutLock):
```
Shrinjay, line 11: Naming is a bit confusing, geolocation_locator_worker(...) would be better

modules/geolocation/geolocationWorker.py
```diff
@@ -0,0 +1,89 @@
...
+
+        if (exit_requested(exitRequest)):
+            break
+
+
+    print("Stop Geolocation")
+    return
+
+
+def geolocation_output(pause, exitRequest, pipelineIn, pipelineOut, pipelineInLock):
```
Shrinjay, line 46: As above, would prefer another function name like geolocation_output_worker(...)

modules/geolocation/geolocationWorker.py
```diff
@@ -0,0 +1,89 @@
...
+
+        # Something has gone wrong, skip
+        if (not ret):
+            continue
+
+        # External lock required as geolocationOutput will dump the queue
+        pipelineOutLock.acquire()
+        pipelineOut.put(location)
+        pipelineOutLock.release()
+
```
Shrinjay, line 37: Can you clarify what the use case is for a pipeline lock? Do we want to prevent the downstream process (assuming it would be geolocation_output) from getting the location before we put it in the queue? Or do we want it to wait until multiple locations are in the pipeline?

modules/geolocation/geolocationWorker.py
```diff
@@ -0,0 +1,89 @@
...
+        pause.release()
+
+        # Dump the contents of the input queue
+        locations = []
+        pipelineInLock.acquire()
+        while True:
+
+            try:
+
+                location = pipelineIn.get_nowait()
```
Shrinjay, line 63: Note that this will raise an Empty exception if the pipeline is empty. I'm guessing the goal with this is to prevent the queue from ever being empty using the pipelineLock? Then get and get_nowait would have no real difference, the only difference occurs if the queue is empty and if we do intend for there to be a case where the queue is intentionally empty, we should have exception handling for the Empty exception. If not, then I don't know why get_nowait() is being used.

modules/geolocation/geolocationWorker.py
```diff
@@ -0,0 +1,89 @@
...
+                # TODO verify this
```
Shrinjay, line 64: I'll maintain the same pattern for returns on run_locator as you have with run_output (return state, returnValue).

modules/geolocation/geolocationWorker.py
```diff
@@ -0,0 +1,89 @@
...
+                if (not location):
```
Shrinjay, line 65: I'll maintain the same pattern as you have for verifying that the pipeline value is valid, I'll just return state, returnValue from run_locator().

modules/geolocation/geolocationWorker.py
```diff
@@ -0,0 +1,89 @@
...
+
+    while True:
+
+        pause.acquire()
+        pause.release()
+
+        # Pixel coordinates of tents and plane data
+        coordinates = pipelineIn.get()
+        # TODO verify this
+        if (not coordinates):
```
Shrinjay, line 24: This one's an interesting case, for now assume an empty list or some other falsey value, I'll have to see how our targetAcquisition module handles exceptions and invalid output.

Shrinjay (changes requested):
Looks good to me overall, few nitpicks, few clarifications.

modules/geolocation/geolocationWorker.py
Xierumeng: In a normal queue, processes take turns putting in/taking out items, while in this case, "taking out items" is grabbing the entire queue and shaking everything out.

modules/geolocation/geolocationWorker.py
Xierumeng: No, we're dumping the entire queue and we don't want to block on empty.

Shrinjay: Got it, in that case you can either check if its empty or just have a try catch block, just because this will throw an error once the entire queue has been extracted and the last one is left.

modules/geolocation/geolocationWorker.py
Shrinjay: Got it

modules/geolocation/geolocationWorker.py
Shrinjay: Oh lmao already done, my bad

modules/geolocation/geolocationWorker.py
```diff
@@ -0,0 +1,92 @@
...
+
+        ret, location = locator.run_locator(coordinates)
+
+        # Something has gone wrong, skip
+        if (not ret):
+            continue
+
+        # External lock required as geolocation_output_worker will dump the queue (without using the internal one)
+        pipelineOutLock.acquire()
+        pipelineOut.put(location)
```
Shrinjay, line 35: I think there's a deadlock condition here? The [put method blocks](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Queue.put) if the pipeline is full (don't think we're determining the size of the queue right now but there should be some kind of innate memory limit, as described in [this issue](https://bugs.python.org/issue8237). Then, this process has the piplineLock, so the output_worker cannot take from the location to empty the pipeline. At the same time, the put method cannot put anything into the pipeline, so it won't release the lock.

Not sure though, not that comfortable with concurrency theory yet. If we do want a work around, checkout [put_nowait ](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Queue.put_nowait), if the queue is full just have it release the lock and continue, maybe?

Shrinjay (commented):
Just one deadlock condition we might need to investigate, but I might be wrong.

Shrinjay (approved):

PR #47, opened 2021-04-27
Command module additions

andrew2002zhao:
- Updated getters and setters to match the exchanged data document
- Updated unit tests to account for new functions

### DIFF
--- a/modules/commandModule/commandModule.py
+++ b/modules/commandModule/commandModule.py
@@ -285,10 +285,68 @@ class CommandModule:
 
         return gpsCoordinates
 
+    def get_editing_flight_path_error_code(self) -> int:
+        self.__read_from_pogi_file()
+        error_code = self.__pogiData["editingFlightPathErrorCode"]
+        if(error_code is None):
+            self.__logger.error("editingFlightPathErrorCode must be an int and not None.")
+            return None
+        if type(error_code is not int):
+            self.__logger.error("{} in editingFlightPathErrorCode is not an int.".format(type(error_code)))
+            return None
+
+        return error_code
+
+    def get_flight_path_following_error_code(self) -> int:
+        self.__read_from_pogi_file()
+        error_code = self.__pogiData["flightPathFollowingErrorCode"]
+        if(error_code is None):
+            self.__logger.error("flightPathFollowingErrorCode must be an int and not None.")
+            return None
+        if type(error_code is not int):
+            self.__logger.error("{} in flightPathFollowingErrorCode is not an int.".format(type(error_code)))
+            return None
+
+        return error_code
+
+    def get_current_waypoint_id(self) -> int:
+        self.__read_from_pogi_file()
+        waypoint_id = self.__pogiData["currentWaypointId"]
+        if(waypoint_id is None):
+            self.__logger.error("currentWaypointId must be an int and not None.")
+            return None
+        if type(waypoint_id is not int):
+            self.__logger.error("{} in currentWaypointId is not an int.".format(type(waypoint_id)))
+            return None
+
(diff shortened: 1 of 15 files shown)


modules/commandModule/commandModule.py
```diff
@@ -404,6 +465,248 @@ def set_begin_takeoff(self, beginTakeoff: bool):
...
+
+        """
+        if waypointModifyFlightPathCommand is None:
+            self.__logger.error("waypointModifyFlightPathCommand must be an int and not None.")
+            return None
+        if type(waypointModifyFlightPathCommand) is not int:
+            self.__logger.error("waypointModifyFlightPathCommand must be an int and not {}.".format(type(waypointModifyFlightPathCommand)))
+            return None
+        self.__pigoData.update({"waypointModifyFlightPathCommand": waypointModifyFlightPathCommand})
+        self.__write_to_pigo_file()
```
Shrinjay, line 500: Something to think about for later, we keep rewriting the entire pigo file when really we should just be updating it.

modules/commandModule/commandModule.py
```diff
@@ -404,6 +465,248 @@ def set_begin_takeoff(self, beginTakeoff: bool):
...
+        Parameters
+        ----------
+        flightPathModifyPrevId: int
+
+        """
+        if flightPathModifyPrevId is None:
+            self.__logger.error("flightPathModifyPrevId must be an int and not None.")
+            return None
+        if type(flightPathModifyPrevId) is not int:
+            self.__logger.error("flightPathModifyPrevId must be an int and not {}.".format(
```
Shrinjay, line 566: Use f strings here, I'm not going to make you replace eveything in this PR but just for future note, .format is Python2-ish iirc, f"...{xyz}..." is more modern and just better (unless you're writing DB queries, but that's another story)

modules/commandModule/commandModule.py
```diff
@@ -404,6 +465,248 @@ def set_begin_takeoff(self, beginTakeoff: bool):
...
+    
+    def set_holding_turn_radius(self, holdingTurnRadius: int):
+        """
+
+        Parameters
+        ----------
+        holdingTurnRadius: int
+
+        """
+        if holdingTurnRadius is None:
```
Shrinjay, line 616: Looking at all this repeated code makes me wonder if we can implement a builder pattern of sorts for this, again not for this PR but for future refactoring. As in have a class CommandBuilder(variableName) with methods buildTypeValidator(expectedType), and buildTypeValidator(...) returns a function called validate{variableName}(variable) that validates the type and makes the appropriate logs.

modules/commandModule/tests/testLogs/12.27.34_2021_04_04
```diff
@@ -1,80 +0,0 @@
-test_value_error_if_get_airspeed_equals_none (commandModule.tests.testCases.testReadNullFromPogi.TestReadingNullFromPOGIFiles) ... ok
```
Shrinjay, line 1: Okay unfortunately I am going to have to ask for some changes here, gitignore the testLogs folder, we don't want this committed to the codebase.

modules/commandModule/tests/testLogs/debug.log
```diff
@@ -1,199 +1,485 @@
-12:27:34,703 root INFO ['ERROR:commandModule.commandModule:currentAirspeed not found in the POGI json file.']
```
Shrinjay, line 1: Same as above, please gitignore this folder

Shrinjay (changes requested):
Just need to gitignore the testLogs folder, I also left some comments we should revisit later for refactoring this, but not needed now.

PR #48, opened 2021-04-28
Added searchWorker.py

thrvg:


modules/search/searchWorker.py
```diff
@@ -0,0 +1,16 @@
...
+
+def searchWorker(pause, exitRequest, pipelineIn, pipelineOut):
+  
+    print("Start Search")
+    
+    search = Search()
+    
+    pause.acquire()
+    
+    values = pipelineIn.get()
```
Shrinjay, line 12: Can you put a comment here that outlines what values is going to look like. Just put the following shape:
```
{
  tentGPS: value,
  planeGPS: value,
  angle: value
}
```

Shrinjay (changes requested):

Shrinjay (approved):
Works for now, since we'll be rewriting parts of this logic anyways

PR #49, opened 2021-05-01
pylon detection module

rayjinghaolei:
implemented pylon detection, please review


modules/targetAcquisition/pylonAcquisition.py
```diff
@@ -0,0 +1,96 @@
...
+        # draw out the bounding boxes
+        for (topLeft, botRight) in self.bbox:
+            cv2.rectangle(self.currentFrame, topLeft, botRight, (0, 0, 255), 2)
+        cv2.imshow('img', self.currentFrame)
+        cv2.waitKey(0) 
+        cv2.destroyAllWindows()
+
+
+# Testing
+tracker = PylonAcquisition()
```
Shrinjay, line 94: This is a good idea but please put this in a separate file. I recommend using pytest, create a /tests/ folder at the root directory and put test scripts in there.

modules/targetAcquisition/pylonAcquisition.py
```diff
@@ -0,0 +1,96 @@
...
+from time import sleep
+
+from tensorflow.compat.v1 import ConfigProto
+from tensorflow.compat.v1 import InteractiveSession
+
+config = ConfigProto()
+config.gpu_options.allow_growth = True
+session = InteractiveSession(config=config)
+
+class PylonAcquisition:
```
Shrinjay, line 15: Only nitpick, please rename this class and the file to targetAcquisition just to keep it all the same.

Shrinjay (changes requested):

modules/targetAcquisition/pylonAcquisition.py
rayjinghaolei: should I just delete the old targetAcquisition and replace it with this one?

modules/targetAcquisition/pylonAcquisition.py
rayjinghaolei: can you show me how it works when we meet tonight?

modules/targetAcquisition/pylonAcquisition.py
Shrinjay: Yes

Shrinjay (approved):

PR #52, opened 2021-05-16
Merge image with telemetry

Wenfei134:
added class and worker function to handle merging image and telemetry pipelines by matching images to the nearest telemetry.

### DIFF
--- a/modules/mergeImageWithTelemetry/mergeImageWithTelemetry.py
+++ b/modules/mergeImageWithTelemetry/mergeImageWithTelemetry.py
@@ -1,0 +1,82 @@
+from modules/timestamp/Timestamp import Timestamp
+
+import datetime
+import typing
+
+class MergeImageWithTelemetry:
+    """
+    Merges image pipeline with telemetry pipeline
+    ...
+
+    Attributes
+    ----------
+    curData : Timestamp 
+        the current telemetry data 
+    start : bool
+        whether curData is None or not (whether get_closest_telemetry has been called before)
+    telemetryData : list<Timestamp>
+        list of the current telemetry data and their times in the form of Timestamp objects
+
+    Methods
+    -------
+    __init__()
+        initiallizes attributes
+    put_back(newTelemetryData: Timestamp)
+        places given telemetry data at the back of the telemetryData list
+    get_closest_telemetry(imageTimestamp: datetime.datetime)
+        finds the telemetry with timestamp closest to the input datetime
+    """
+
+    def __init__(self): 
+        self.curData = None
+        self.start = True
+        self.telemetryData = []
+
+    def put_back(self, newTelemetryData: typing.type[Timestamp]):
+        self.telemetryData.append(newTelemetryData)
+
+    def get_closest_telemetry(self, imageTimestamp: datetime.datetime): 
+        """
+        finds the telemetry with timestamp closest to the input datetime
+        assumes that TelemetryData is sorted by increasing timestamp
+
+        Parameters 
+        ----------
+        imageTimestamp : datetime.datetime 
+            The time that the telemetry data should match with
+
+        Returns 
+        -------
+        [success, telemetry]
+
+        success : bool 
+            whether such a telemetry timestamp could be found
+        telemetry : Timestamp
+            the closest time and the corresponding telemetry data in a Timestamp object
+        """
+
+        if len(self.telemetryData) == 0:
+            if self.start: 
+                return False, None
+            return True, self.curData
+        
+        if self.start: 
+            self.curData = self.telemetryData.pop(0)
+            self.start = False
+        
+        timeDelta = abs(imageTimestamp - self.curData.timestamp)
+
+        while True:
+            if len(self.telemetryData) == 0:
+                return True, self.curData
+            
+            nextTelemetry = self.telemetryData.pop(0)
+            nextTimeDelta = abs(nextTelemetry.timestamp - imageTimestamp)
+            
+            if nextTimeDelta > timeDelta:
+                ret = self.curData 
+                self.curData = nextTelemetry
+                return True, ret
+            else: 
+                self.curData = nextTelemetry
+                timeDelta = nextTimeDelta
--- a/modules/mergeImageWithTelemetry/mergeImageWithTelemetry.py
+++ b/modules/mergeImageWithTelemetry/mergeImageWithTelemetry.py
@@ -8,6 +8,8 @@ import typing
 class MergeImageWithTelemetry:
     """
     Merges image pipeline with telemetry pipeline
+
+    The imports used by this class may require python 3.8 or greater
     ...
 
     Attributes
@@ -25,6 +27,8 @@ class MergeImageWithTelemetry:
         whether a new image should be inputted to be matched 
             - old image has already matched a telemetry timestamp or
             - pipeline is just starting and there is no image yet
+    set_image(image: Timestamp): 
+        sets the input to be the current image
     put_back_telemetry(newTelemetryData: Timestamp)
         places given telemetry data at the back of the telemetryData list
     get_closest_telemetry()
@@ -49,13 +53,6 @@ class MergeImageWithTelemetry:
         finds the telemetry with closest timestamp to the image timestamp
         assumes that TelemetryData is sorted by increasing timestamp
 
-        Parameters 
-        ----------
-        imageTimestamp : datetime.datetime 
-            The time that the telemetry data should match with
-        imageData : npm.ArrayLike
-            The image data corresponding to the timestamp
-
         Returns 
         -------
         [success, mergedData]
@@ -78,7 +75,7 @@ class MergeImageWithTelemetry:
         timeDelta = abs(imageTimestamp - oldestTelemetry.timestamp)
 
         while True:
-            # if we run out of telemetry data should we match with the last one or should we return false 
+            # if we run out of telemetry data, unget the current telemetry and just return false 
             if len(self.telemetryData) == 0:
                 self.telemetryData.append(oldestTelemetry)
                 return False, None
@@ -87,6 +84,7 @@ class MergeImageWithTelemetry:
             nextTimeDelta = abs(nextTelemetry.timestamp - imageTimestamp)
             
             if nextTimeDelta > timeDelta:
+                # un_get the next telemetry
                 self.telemetryData.insert(0, nextTelemetry)
                 self.image = None
                 return True, MergedData(imageData, oldestTelemetry.data)
--- a/modules/mergeImageWithTelemetry/mergeImageWithTelemetry.py
+++ b/modules/mergeImageWithTelemetry/mergeImageWithTelemetry.py
@@ -34,7 +34,7 @@ class MergeImageWithTelemetry:
         self.start = True
         self.telemetryData = []
 
-    def put_back(self, newTelemetryData: typing.type[Timestamp]):
+    def put_back(self, newTelemetryData: Timestamp):
         self.telemetryData.append(newTelemetryData)
 
     def merge_with_closest_telemetry(self, imageTimestamp: datetime.datetime, imageData : npt.ArrayLike ): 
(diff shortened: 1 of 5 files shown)


modules/mergeImageWithTelemetry/mergeImageWithTelemetryWorker.py
```diff
@@ -0,0 +1,45 @@
...
+        pause.release()
+        
+        if shouldGetImage:
+            curImage = imagePipelineIn.get()
+
+        if curImage is None:
+            continue
+
+        # put current elements of the telemetry pipeline into the queue
+        while not telemetryPipelineIn.empty():
```
Shrinjay, line 23: This has the potential to be an infinite loop, the telemetryPipeline will keep getting filled faster than this can pull, instead, each run should just get the latest data in the telemetry pipeline. Essentially, each time the **parent** loop runs, it should get the latest telemetry, add it to the telemtryData array, get the current image,and matches the current image with the correct telemetry Data. 

Because the decklinksrc program will almost certainly run slower than the command module, we know that whatever image we have at that point is certainly older than the item at the head of the telemetry pipeline (assuming LIFO). Because we add the newest data to the tail of the pipeline, we logically know that any data older than the item at the head of the telemetry pipeline is already in the telemetryData array. We know that the telemetry data that matches the image will be older than the item at the head of the telemtry pipeline, therefore we can deduce that by implementing the change in the comment above, the data with a timestamp closest to the current image is in telemetryData somewhere. 

I wish I could write this in clean mathematical language but not possible here, we can discuss this further as well.

modules/mergeImageWithTelemetry/mergeImageWithTelemetryWorker.py
```diff
@@ -0,0 +1,45 @@
...
+        if shouldGetImage:
+            curImage = imagePipelineIn.get()
+
+        if curImage is None:
+            continue
+
+        # put current elements of the telemetry pipeline into the queue
+        while not telemetryPipelineIn.empty(): 
+        
+            if not exitRequest.empty():
```
Shrinjay, line 25: The exitRequest and puase logic from line 25-29 doesn't need to be here, we don't need to do this every single time we get from a pipeline.

modules/mergeImageWithTelemetry/mergeImageWithTelemetryWorker.py
```diff
@@ -0,0 +1,45 @@
...
+    curImage = None
+
+    while True: 
+        if not exitRequest.empty():
+            return
+        
+        pause.acquire()
+        pause.release()
+        
+        if shouldGetImage:
```
Shrinjay, line 16: I think we should get the telemetry pipeline **first**, add it to the telemetryData array, then if an image isn't available, we still have the telemetry data. This means that, assuming the image pipeline producer runs slower than the telemetry pipeline producer, we should know that the item at the head of the telemetry pipeline (assuming LIFO) is newer than the item at the head of the image pipeline. Therefore, we know that the telemetryData with the closest timestamp to the image should be somewhere in the telemetryData array, not stuck in the telemetry pipeline.

modules/mergeImageWithTelemetry/mergeImageWithTelemetryWorker.py
```diff
@@ -0,0 +1,45 @@
...
+
+            telemetryData = telemetryPipelineIn.get()
+
+            mergeImageWithTelemetry.put_back(telemetryData)
+
+        [success, mergedData] = mergeImageWithTelemetry.merge_with_closest_telemetry(curImage.timestamp, curImage.data)
+
+        if success: 
+            pipelineOut.put(mergedData)
+            shouldGetImage = True # curImage has been matched with a telemetry
```
Shrinjay, line 39: Good call, didn't consider this

Shrinjay (changes requested):
The fusion logic itself is fine, a bit confusing at first read but I think your implementation is the most efficient one memory and time wise. However the worker process needs some work (pun intended). Multiprocessing is a difficult topic and when we think of this stuff we have to essentially construct a proof to show that a certain thing is guaranteed to happen before another, so its hard to communicate via comments, I tried my best. We can talk further next meeting.

modules/mergeImageWithTelemetry/mergeImageWithTelemetryWorker.py
Shrinjay, line 23: Use locking logic as seen in geolocationWorker, acquire lock, dump queue, release lock

modules/mergeImageWithTelemetry/mergeImageWithTelemetryWorker.py
```diff
@@ -0,0 +1,45 @@
+from modules.mergeImageWithTelemetry.mergeImageWithTelemetry import MergeImageWithTelemetry
+
+def pipelineMergeWorker(pause, exitRequest, imagePipelineIn, telemetryPipelineIn, pipelineOut):
+    print("start pipeline merge")
+    mergeImageWithTelemetry = MergeImageWithTelemetry()
+    shouldGetImage = True
+    curImage = None
```
Xierumeng, line 7: I would prefer the image to be stored inside the class object.

modules/mergeImageWithTelemetry/mergeImageWithTelemetry.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def __init__(self): 
+        self.curData = None
+        self.start = True
+        self.telemetryData = []
+
+    def put_back(self, newTelemetryData: typing.type[Timestamp]):
+        self.telemetryData.append(newTelemetryData)
+
+    def merge_with_closest_telemetry(self, imageTimestamp: datetime.datetime, imageData : npt.ArrayLike ):
```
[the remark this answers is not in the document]
Xierumeng: It's fine. It's pretty self-explanatory in the current form (takes in an image timestamp and an image and merges it with the closest telemetry (which at first glance is assumed to also contain a timestamp, which it does)).

modules/mergeImageWithTelemetry/mergeImageWithTelemetry.py
```diff
@@ -0,0 +1,84 @@
...
+
+class MergeImageWithTelemetry:
+    """
+    Merges image pipeline with telemetry pipeline
+    ...
+
+    Attributes
+    ----------
+    curData : Timestamp 
+        the current telemetry data
```
Xierumeng, line 16: By "current" do you mean oldest in the list?

modules/mergeImageWithTelemetry/mergeImageWithTelemetry.py
```diff
@@ -0,0 +1,84 @@
...
+        success : bool 
+            whether such a telemetry timestamp could be found
+        telemetry : Timestamp
+            the closest time and the corresponding telemetry data in a Timestamp object
+        """
+
+        if len(self.telemetryData) == 0:
+            if self.start: 
+                return False, None
+            return True, self.curData
```
Xierumeng, line 63: There is the case that the old image is matched, no new telemetry data comes in, and then a new image comes in and is matched with the exact same telemetry. What should probably happen is to hold onto the image and try to get new telemetry data instead.

Of course, this case should never happen (otherwise something has gone terribly wrong), but it's still possible (maybe the plane experiences a hiccup).

modules/mergeImageWithTelemetry/mergeImageWithTelemetry.py
```diff
@@ -0,0 +1,84 @@
...
+class MergeImageWithTelemetry:
+    """
+    Merges image pipeline with telemetry pipeline
+    ...
+
+    Attributes
+    ----------
+    curData : Timestamp 
+        the current telemetry data 
+    start : bool
```
Xierumeng, line 17: Use a descriptive modifier (e.g. hasStarted, hasNotStarted, isStarted, etc.). In general all boolean variables should have them.

modules/mergeImageWithTelemetry/mergeImageWithTelemetryWorker.py
Xierumeng: I believe this is an attempt to continuously get telemetry data to match to the image. Unfortunately, once the image is matched, it will still continuously get more telemetry data. There is a possibility of spinning forever as the process putting telemetry into the pipeline can do so at an equal or faster rate than this process can extract it.

Additionally, it is relatively expensive to dribble the telemetry into the class object.

modules/mergeImageWithTelemetry/mergeImageWithTelemetryWorker.py
```diff
@@ -0,0 +1,45 @@
...
+
+        [success, mergedData] = mergeImageWithTelemetry.merge_with_closest_telemetry(curImage.timestamp, curImage.data)
+
+        if success: 
+            pipelineOut.put(mergedData)
+            shouldGetImage = True # curImage has been matched with a telemetry
+        else: 
+            shouldGetImage = False # curImage has not been matched with a telemetry and we should keep it as the current image
+
+
```
Xierumeng, line 43: Formatting nitpick: Remove these extra lines, there should only be one empty line at the end.

modules/mergeImageWithTelemetry/mergeImageWithTelemetryWorker.py
```diff
@@ -0,0 +1,45 @@
...
+        
+
```
Xierumeng, line 45: Formatting nitpick: Remove these extra lines, there should only be one empty line at the end.

modules/mergeImageWithTelemetry/mergeImageWithTelemetry.py
Xierumeng: There seems to be a lot of unnecessary code dealing with `curData` (whether it contains valid data at the start, updating it when it's matched, etc.). Since there are already checks for `len(telemetryData) == 0`, why not use `telemetryData[0]` directly?

Xierumeng:
Also, where are the unit tests?

modules/mergeImageWithTelemetry/mergeImageWithTelemetryWorker.py
```diff
@@ -0,0 +1,39 @@
...
+    while True: 
+        if not exitRequest.empty():
+            print("end pipeline merge")
+            return
+        
+        pause.acquire()
+        pause.release()
+
+        # put current elements of the telemetry pipeline into the queue
+        telemetryPipelineIn.acquire()
```
Xierumeng, line 17: A separate lock is supposed to be passed into the worker function. The queue/pipeline itself cannot be locked like this.

modules/mergeImageWithTelemetry/mergeImageWithTelemetryWorker.py
```diff
@@ -0,0 +1,39 @@
...
+            try: 
+                telemetryData = telemetryPipelineIn.get_noawait()
+                mergeImageWithTelemetry.put_back_telemetry(telemetryData)
+            except: 
+                break
+        telemetryPipelineIn.release()
+        
+        if mergeImageWithTelemetry.should_get_image():
+            try: 
+                curImage = imagePipelineIn.get_noawait()
```
Xierumeng, line 28: There's no need to constantly try to get an image, if there's no image right now it's fine to wait patiently for one.

Basically right now the worker is going to be spinning a lot, which is wasteful.

Xierumeng: So something like:
```
if mergeImageWithTelemetry.should_get_image():
    curImage = imagePipelineIn.get()  # We can wait for an image, no rush
```
is fine.

tests/merge_image_with_telemetry_test.py
```diff
@@ -0,0 +1,110 @@
...
+
+    telemetry = Timestamp(["telemetry"], now)
+    merger.put_back_telemetry(telemetry)
+
+    [success, output] = merger.get_closest_telemetry()
+    assert [success, output] == [False, None]
+    # made it to the end, the telemetry pipeline is waiting
+
+if __name__ == "__main__": 
+    print("testing empty")
```
Xierumeng, line 93: Uh... there has to be a better way of doing this. At the very least, move the print statements into the functions themselves.

tests/merge_image_with_telemetry_test.py
```diff
@@ -0,0 +1,110 @@
...
+    print("****************** PASSED ************************")
+    print("testing first")
+    first()
+    print("****************** PASSED ************************")
+    print("testing middle")
+    middle()
+    print("****************** PASSED ************************")
+    print("testing last")
+    last()
+    print("****************** PASSED ************************")
```
Xierumeng, line 110: Formatting nitpick: Single empty line at end of file.

PR #56, opened 2021-05-27
Add worker function for POGI data

ksisjaya:
- Add worker function to pipe POGI data from POGI json file
- Add util function to read from POGI json file
- Note: util function explicitly calls commandModule getter for every field to be defined in pogiData dict

### DIFF
--- a/modules/commandModule/commandFns.py
+++ b/modules/commandModule/commandFns.py
@@ -104,7 +104,7 @@ def read_pogi(POGI_DIR="") -> tuple:
     'gpsCoordinates' : command.get_gps_coordinates(),
     }
 
-    latestJsonDirectory = ".\\latest_pogi.json"
+    latestJsonDirectory = os.path.join(os.getcwd(), "modules", "commandModule", "latest_pogi.json")
     isChanged = json_changed(latestJsonDirectory, pogiData)
     
     # if pogi has changed, then write new data to latest_pogi.json
--- a/modules/commandModule/commandFns.py
+++ b/modules/commandModule/commandFns.py
@@ -1,6 +1,7 @@
 from modules.commandModule.commandModule import CommandModule
 
 PIGO_DIR = ""
+POGI_DIR = ""
 
 def write_pigo(newPigo):
 #Boolean to check if any changes were made
@@ -32,3 +33,29 @@ def write_pigo(newPigo):
 #if nothing has changed, return the dictionary that was passed in
     else:
         return
+
+def read_pogi() -> dict:
+
+    """
+    Reads all data currently in the POGI file and returns as a dict
+
+    Returns
+    -------
+    dict:
+        contains pogiData as specified in commandModule
+    """
+
+    command = CommandModule(pigoFileDirectory=PIGO_DIR, pogiFileDirectory=POGI_DIR)
+
+    # store all current POGI data into a dict
+    # note: the keys included in the dict are those which are currently valid and working in commandModule
+    #       (some fields are defined in the wiki but are currently commented out)
+    pogiData =  { 'currentAltitude' : command.get_current_altitude(),
+                  'currentAirspeed' : command.get_current_airspeed(),
+                  'isLanded' : command.get_is_landed(),
+                  'eulerAnglesOfCamera' : command.get_euler_angles_of_camera(),
+                  'eulerAnglesOfPlane' : command.get_euler_angles_of_plane(),
+                  'gpsCoordinates' : command.get_gps_coordinates(),
+                }
+    
+    return pogiData
--- a/modules/commandModule/commandFns.py
+++ b/modules/commandModule/commandFns.py
@@ -1,7 +1,47 @@
+import json
+import os
 from modules.commandModule.commandModule import CommandModule
 
 PIGO_DIR = ""
 POGI_DIR = ""
+
+def json_changed(latestJsonDirectory, currentDict) -> bool:
+
+    """
+    Compares json values to a dict
+
+    Parameters
+    ----------
+    latestJsonDirectory: str
+        filepath to the latest json file to compare to
+    currentDict: dict
+        dictionary containing values to compare with
+
+    Returns
+    -------
+    bool:
+        True if json file and the dict do not match, False otherwise
+    """
+
+    # if latest pogi data file doesn't exist return true
+    if not os.path.isfile(latestJsonDirectory):
+        return True
+
+    # otherwise, get the latest json dict and compare
+    with open(latestJsonDirectory, 'r') as f:
+        latestDict = json.load(f)
+
+    # if current and latest pogi data doesnt match size, then something changed
+    if len(latestDict.keys()) != len(currentDict.keys()):
+        return True
+
+    # if keys in current pogi don't exist in latest pogi or if the values for the keys
+    # don't match, then something changed
+    for key in currentDict:
+        if not (key in latestDict and latestDict[key] == currentDict[key]):
+            return True
+    
+    return False
 
 def write_pigo(newPigo):
 #Boolean to check if any changes were made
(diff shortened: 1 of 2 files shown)


modules/commandModule/commandWorker_flight.py
```diff
@@ -15,4 +19,34 @@ def flight_command_worker(pipelineIn, pipelineOut):
...
+    """
+
+    print("Reading POGI data")
+
+    while True:
+        pause.acquire()
+        pause.release()
+
+        # get the pogi data
+        pogiData = read_pogi()
```
Shrinjay, line 44: I'm going to slightly expand the scope of this ticket, I apologize for that, but I think there's a better way to do this. 

Essentially, we're going to need to resuse this logic in every worker function (one for flight, one for taxi, one for search). Therefore, we should pull it out into its own function. Instead of having this worker function, write something called `pogi_subworker(pipelineOut, POGI_DIR)`. This should be a function that encompasses line 43 - 49, just the logic to read the latest pogi data, compare it against the last one, and put it in the pipeline. 

Then, we should be able to call this in the other worker functions, so each time those worker functions run, they first perform whatever activities they're supposed to do, then they update their pipelineOut with the latest pogi data.

modules/commandModule/commandWorker_flight.py
```diff
@@ -1,4 +1,8 @@
-from modules.commandModule.commandFns import write_pigo
+from modules.commandModule.commandModule import CommandModule
+from modules.commandModule.commandFns import write_pigo, read_pogi
+
+POGI_DIR = ""
+PIGO_DIR = ""
 
 def flight_command_worker(pipelineIn, pipelineOut):
 	# pipelineIn gives [[x,y], [range]]
```
Shrinjay, line 8: Out of scope but could you please put the logic here (lines 4-18) inside a while loop. Then implement the changes in my following comment and call `pogi_subworker(...)` here, passing in the pipeline out and directory.

modules/commandModule/commandFns.py
```diff
@@ -31,4 +32,30 @@ def write_pigo(newPigo):
 	return
 #if nothing has changed, return the dictionary that was passed in
     else:
-        return
\ No newline at end of file
+        return
+
+def read_pogi() -> dict:
```
Shrinjay, line 37: It would be really useful if we could have a way to tell if the pogi has been updated, since we don't want to shove redundant data into our pipelines. Here's my idea for implementing this:

1) Once you've run everything in lines 53-60, save that pogi as a json called last_pogi
2) Everytime this function is run, first have it check if a last_pogi.json exists. If it does, load it, and using a similar algorithm to what is used in write_pogi to check if anything has been changed. 
3) If nothing has been changed, return `False, pogiData`, if something has been changed return `True, pogiData`.

modules/commandModule/commandWorker_flight.py
Shrinjay, line 44: Using the changes I mentioned above, `read_pogi()` should return a tuple where the first entry is a boolean representing if the pogi has been changed, and the second is the pogi data itself. Change this line to `changed, pogi_data = read_pogi()`.

modules/commandModule/commandWorker_flight.py
```diff
@@ -15,4 +19,34 @@ def flight_command_worker(pipelineIn, pipelineOut):
...
+        pause.acquire()
+        pause.release()
+
+        # get the pogi data
+        pogiData = read_pogi()
+
+        # output to pipeline
+        # note: no null checking required since CommandModule returns None for fields that are unreadable/inaccesible
+        #       if all fields are None, then pogiData will be a dict with all elements as {'example_key' : None}
+        pipelineOut.put(pogiData)
```
Shrinjay, line 49: Using what I mentioned above, add an if statement here that checks if `changed` is `True` before putting the pogiData in the pipeline.

modules/commandModule/commandFns.py
```diff
@@ -31,4 +32,30 @@ def write_pigo(newPigo):
...
+    dict:
+        contains pogiData as specified in commandModule
+    """
+
+    command = CommandModule(pigoFileDirectory=PIGO_DIR, pogiFileDirectory=POGI_DIR)
+
+    # store all current POGI data into a dict
+    # note: the keys included in the dict are those which are currently valid and working in commandModule
+    #       (some fields are defined in the wiki but are currently commented out)
+    pogiData =  { 'currentAltitude' : command.get_current_altitude(),
```
Shrinjay, line 53: Use standard formatting for dictionaries, there should be a newline after the opening bracket, similar to what is done in write_pogi(). Unfortunately the github editor won't let me give you an example. 
`

Shrinjay (changes requested):
Overall your code is pretty good, but I think the fundamental design of this could be tweaked a bit to be more effective.

modules/commandModule/commandWorker_flight.py
ksisjaya: sorry...not too sure what you mean. do you mean put the pogi and pigo logic into one while loop in one worker function? if so, then would you like me to write up a `pigo_subworker(...)` as well?

modules/commandModule/commandFns.py
```diff
@@ -31,4 +72,44 @@ def write_pigo(newPigo):
...
+    pogiData =  {
+    'currentAltitude' : command.get_current_altitude(),
+    'currentAirspeed' : command.get_current_airspeed(),
+    'isLanded' : command.get_is_landed(),
+    'eulerAnglesOfCamera' : command.get_euler_angles_of_camera(),
+    'eulerAnglesOfPlane' : command.get_euler_angles_of_plane(),
+    'gpsCoordinates' : command.get_gps_coordinates(),
+    }
+
+    latestJsonDirectory = ".\\latest_pogi.json"
```
manumanuk, line 107: Are you trying to access a path from the current directory? If so, use `os.path.join(os.getcwd(), "latest_pogi.json")`. Also note that when the program runs, os.getcwd() will always be the root folder where main.py is running, so you may need to navigate to an additional folder.

modules/commandModule/commandWorker_flight.py
```diff
@@ -1,18 +1,56 @@
...
+        
+        # PIGO Logic
+        # pipelineIn gives [[x,y], [range]]
+        if pipelineIn.empty:
+            continue
+        
+        data = pipelineIn.get()
+        # Cache data here
+        newPigo = {
+            'gpsCoordinates': [x, y]
```
manumanuk, line 47: I think you meant to say data[0]

manumanuk: Instead of [x, y]

modules/commandModule/commandWorker_flight.py
```diff
@@ -1,18 +1,56 @@
...
+    """
+
+    while True:
+
+        pause.acquire()
+        pause.release()
+        
+        # PIGO Logic
+        # pipelineIn gives [[x,y], [range]]
+        if pipelineIn.empty:
```
manumanuk, line 41: This means if the in pipeline is empty we will have to wait to get in PIGO data. The POGI logic should be moved above the PIGO logic, so that we make sure we're not delaying the flow of telemetry data in.

modules/commandModule/commandWorker_flight.py
```diff
@@ -1,18 +1,56 @@
...
+            continue
+        
+        data = pipelineIn.get()
```
manumanuk, line 44: The only slight issue with using one worker for both POGI and PIGO is that if there is no incoming telemetry, we'll be stuck at this line, and won't be able to send any commands to the plane until we get a telemetry reading, which isn't bad I suppose because:
1. We'll most likely always be getting a telemetry reading
2. If we're not, we probably won't be able to uplink anyway

But I need a thumbs up from @Shrinjay on this, because otherwise we'll have to separate POGI and PIGO into their own separate workers.

manumanuk: If Shrinjay gives a thumbs up, no need to make a new worker

manumanuk (changes requested):
Few small changes required. Once those are done, wait for Shrinjay to confirm the one comment about separating the workers, and he can merge it in or decide to ask you to refactor the worker code.

manumanuk:
Looks good to me, just make sure to pull from main and resolve merge conflicts

manumanuk (approved):

PR #60, opened 2021-06-11
Added stop command to Taxi Worker

MichaelAiz:


modules/targetAcquisition/taxi/TaxiWorker.py
```diff
@@ -9,9 +9,8 @@ def taxi_worker(pause, exitRequest, pipelineIn, pipelineOut):
         pause.release()
         
         frame = pipelineIn.get()
-        taxi.main(frame)
-
-        pipelineOut.put()
+        if taxi.main(frame).latestDistance == 0:
+            pipelineOut.put(taxi.main(frame))
```
Shrinjay, line 13: Minor thing, don't call main over and over again. Call taxi.main(frame) and save it to a variable like ```command = taxi.main(frame)``` use that variable instead

modules/targetAcquisition/taxi/TaxiWorker.py
```diff
@@ -9,9 +9,8 @@ def taxi_worker(pause, exitRequest, pipelineIn, pipelineOut):
         pause.release()
         
         frame = pipelineIn.get()
-        taxi.main(frame)
-
-        pipelineOut.put()
+        if taxi.main(frame).latestDistance == 0:
```
Shrinjay, line 12: First check that the return from taxi.main has a property called latestDistance, use OR.

Xierumeng, line 12: Data validation: check incomging and outgoing data.

modules/targetAcquisition/taxi/TaxiWorker.py
```diff
@@ -9,9 +9,8 @@ def taxi_worker(pause, exitRequest, pipelineIn, pipelineOut):
         pause.release()
         
         frame = pipelineIn.get()
```
Shrinjay, line 11: Validate that frame is not None, if it is, just continue

Shrinjay (changes requested):

PR #62, opened 2021-06-16
Flight Driver

manumanuk:
# Purpose
To update `main.py` with complete pipeline & worker implementations

# New Changes
- Created pipelines corresponding to data going in and out of different modules
- Added pause & exitrequest features to command module: **does this need to be removed, since exiting flight should not exit command?**

# Outstanding Changes
- POGI worker logic still undefined
- **What does geolocation_locator_worker do?** There is no method `.run_locator()` on `Geolocation`, it is unclear what the output data is, so right now the output pipeline is just referred to as `geolocationIntermediaryPipeline`

**CRITICAL:**
- There are some major issues with architecture:
     - Fusion module should be merging bbox data with telemetry, not raw images.
     - Either geolocation must interpret the bbox data structure, or targetAcquisition must be changed to output just central pixel coordinates rather than bbox data
     - Geolocation needs to be able to interpret MergedData objects to extract telemetry data
     - Command Module currently expects coordinatePipeline input in the form [[x, y], [range]]. Where is this data structure expected to come from?

### DIFF
--- a/main.py
+++ b/main.py
@@ -3,6 +3,11 @@ import os
 import multiprocessing as mp
 from modules.targetAcquisition.targetAcquisitionWorker import targetAcquisitionWorker
 from modules.decklinksrc.decklinkSrcWorker import decklinkSrcWorker
+from modules.commandModule.commandWorker_flight import flight_command_worker
+from modules.mergeImageWithTelemetry.mergeImageWithTelemetryWorker import pipelineMergeWorker
+from modules.geolocation.geolocationWorker import geolocation_locator_worker, geolocation_output_worker
+
+PIGO_DIRECTORY = ""
 
 # Main process called by command line
 # Main process manages PROGRAMS, programs call submodules for data processing and move data around to achieve a goal.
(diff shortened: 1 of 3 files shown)


Shrinjay:
Responses to each critical concern: 

a) I agree, I think we should be merging bounding box coordinates. However, the image does need to retain the timestamp. Essentially, we need to make some updates. 

decklinkSrc (worker function needs to timestamp frames and output timestamp object) -> targetAcquisition (needs to handle timestamp class, generate a new timestamped bbox cord output) -> fusion (behaviour here should be unchanged). 

b) Good catch! I didn't notice this, let's make a ticket for it. 

c) Yeah, there's a ticket for updating geoloation's worker

d) Should be coming from geolocation_output. 

As for your question about geolocation_run, honestly, I haven't written the code for it yet lmao. Will be done before tomorrow night. For the POGI worker, yeah, PR already in for that, working through it.

manumanuk:
New version ready, please merge PR #66 and PR #56 before reviewing

Shrinjay (approved):

PR #65, opened 2021-06-17
DeckLink

Xierumeng:
To use OBS, stream it to URL (rdp, rtsp, etc.), Change `cv2.VideoCapture()` accept the URL string as input. Source: https://stackoverflow.com/questions/49978705/access-ip-camera-in-python-opencv


manumanuk (approved):
Looks good

PR #68, opened 2021-06-18
Taxipy return

thrvg:
This is a PR with Taxi Worker updates and Stop Command updates. I made another separate PR with just the Taxi Worker updates as well, just in case.

### DIFF
--- a/modules/targetAcquisition/taxi/TaxiWorker.py
+++ b/modules/targetAcquisition/taxi/TaxiWorker.py
@@ -1,6 +1,27 @@
 from modules.targetAcquisition.taxi import Taxi
 
 def taxi_worker(pause, exitRequest, pipelineIn, pipelineOut):
+    """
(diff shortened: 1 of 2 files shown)


modules/targetAcquisition/taxi/TaxiWorker.py
```diff
@@ -9,9 +30,12 @@ def taxi_worker(pause, exitRequest, pipelineIn, pipelineOut):
         pause.release()
         
         frame = pipelineIn.get()
-        taxi.main(frame)
-
```
Shrinjay, line 13: Pls remove all your changes from TaxiWorker, I'd recommend just unstaging it. Reason is that there's another PR I just merged that implements this with validation which is preferable, my apologies for assigning overlapping tasks.

Shrinjay (changes requested):

PR #73, opened 2021-06-22
Fixed Search Implementation, Fixed worker function, wrote tests

Shrinjay:


modules/commandModule/commandWorker_taxi_first.py
```diff
@@ -1,25 +1,27 @@
...
-		return
-	
-	turn_command_data = pipelineIn.get()
-	set_ground_commands(turn_command_data)
+    Returns
+    -------
+    None
+    """
+
+    if pipelineIn.empty():
```
manumanuk, line 22: Should probably just delete this if statement, doing `pipelineIn.get()` on line 26 already handles the empty case for us. I can do this deletion in my own PR and I'll just merge this in for now.

manumanuk (commented):
Good to merge

PR #74, opened 2021-06-22
Merge Taxi Branch into Main

Shrinjay:


modules/targetAcquisition/taxi/Taxi.py
```diff
@@ -0,0 +1,331 @@
...
+        state: string
+            state of object recognition ("BOX", "TRACK", ...)
+        """
+        if state == "BOX" or state == 0:
+            self.state = "BOX"
+
+        elif state == "TRACK" or state == 1:
+            self.state = "TRACK"
+
+            # If switching to track from box for the first time, use the box ID
```
manumanuk, line 176: Random indent

manumanuk: Assuming there was supposed to be an else here?

modules/targetAcquisition/taxi/Taxi.py
```diff
@@ -0,0 +1,331 @@
...
+            else:
+                frameCount = 0
+
+            if frameCount == self.numStableFrames:
+                print("Got the required number of stable frames")
+                self.set_state("TRACK")
+                frameCount = 0
+
+            self.totalWait += 1
+            if self.totalWait > 250:
```
manumanuk, line 245: Double check if this is a good number to stick with: at 15fps, 250 frames is 17 seconds.

modules/targetAcquisition/taxi/Taxi.py
```diff
@@ -0,0 +1,331 @@
...
+    # Constant declaration
+    
+    FOCAL_LENGTH = 24
+    REAL_HEIGHT = 101.6
+    IMAGE_HEIGHT = 1080
+    SENSOR_HEIGHT = 4.63
+    
+    def __init__(self, state="BOX", bbox=[((0, 0), (0, 0))], frame=[], nextUncheckedID=0,
+                 expectedCount=5, numStableFrames=20, distanceFromBox=0,
+                 minDistanceFromBox=0, moveWaitTarget=0, recalibrate=False, lastBbox=[],
```
manumanuk, line 83: Should set some of these values (particularly minDistanceFromBox) to something nonzero

manumanuk (changes requested):
Few important changes to make before a merge can happen. Please fix as soon as possible

modules/targetAcquisition/taxi/Taxi.py
```diff
@@ -0,0 +1,331 @@
...
+                p1 = (int(bbox[0]), int(bbox[1]))
+                p2 = (int(bbox[0] + bbox[2]), int(bbox[1] + bbox[3]))
+                self.bbox = [(p1, p2)]
+                cv2.rectangle(self.frame, p1, p2, (255, 0, 0), 2, 1)
+
+                # While not at half the distance from the box, keep moving forward
+                # if distanceCount < self.distanceFromBox // 2:
+                if self.distanceCount < 500:
+                    # print("While not at half the distance from the box, keep moving forward")
+                    # Issue the next movement command every couple of frames
```
manumanuk, line 266: Again we never actually end up sending a command anywhere in this file. This needs to be fixed

modules/targetAcquisition/taxi/Taxi.py
Shrinjay: No, this is correct, its just not very readable due to the space. this is all tracking logic.

modules/targetAcquisition/taxi/Taxi.py
Shrinjay: 10 seconds sounds like a nicer round number maybe?

requirements.txt
```diff
@@ -18,4 +18,36 @@ torch>=1.7.0
...
 tqdm>=4.41.0
 
-pycocotools>=2.0
\ No newline at end of file
+pycocotools>=2.0
+
+# FROM BOX DETECTECTION: DUPLICATES POSSIBLE
+
+# base ----------------------------------------
+Cython
```
manumanuk, line 26: There are duplicates in this file, (probably) won't cause an issue though. Also I'm not 100% sure if the comments could affect something

modules/targetAcquisition/taxi/Taxi.py
```diff
@@ -260,50 +261,53 @@ def main(self, cap):
...
                 # At the halfway point, do YOLO again to reset the size of the bounding box
                 # KCF and most other tracking algos have constant bbox size even when getting closer, which is a problem for distance calculation
                 else:
                     print("Distance limit reached, time to recalibrate")
-                    distanceCount = 0
+                    self.distanceCount = 0
                     self.recalibrate = True
                     self.lastBbox = self.bbox
                     self.set_state("BOX")
+                    return {'latestDistance': 0.0}
```
manumanuk, line 303: I'm wondering if it might just be better to return `None` in these cases

manumanuk: We already catch that in the worker

manumanuk (commented):
Quick things about imports and return values

modules/targetAcquisition/taxi/TaxiWorker.py
```diff
@@ -1,19 +1,39 @@
...
 
     while True:
         pause.acquire()
         pause.release()
         frame = pipelineIn.get()
-        if (not frame == None):
+        if frame is not None:
             command = taxi.main(frame)
-            if (hasattr(command, latestDistance)) and (command.latestDistance == 0):
+            if ('latestDistance' in command) and (command.latestDistance == 0):
```
Xierumeng, line 35: We're only sending the distance if it's 0?

manumanuk (approved):

PR #76, opened 2021-06-23
taxi-driver

manumanuk:
Taxi driver is complete, there are no issues with data pipelines other than the fact that we still have to fix pogi to give us timestamped data instead (@Wenfei134).

Additional changes that I added along with the taxi driver code:
- Adjusted taxi_command_worker_first to only work with a single parameter and no pipelines (it wasn't necessary, since this won't be multithreaded, it just needs to run once)
- Created a command_taxi_worker_continuous to handle input from taxi pipeline sending commands to Command module

### DIFF
--- a/main.py
+++ b/main.py
@@ -1,13 +1,17 @@
 import argparse
 from datetime import datetime
 import logging
+from modules.targetAcquisition.taxi.TaxiWorker import taxi_worker
 import os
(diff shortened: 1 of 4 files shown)


manumanuk:
**Just added some logic to timestamp pogi data.**

Shrinjay (approved):

PR #81, opened 2021-08-01
Added mock command module

K1VN:


FizzesShark:
Missing the currentAltitude field, and I believe the eulerAnglesOfCamera should be a dictionary rather than an array, and the same with eulerAnglesOfPlane and gpsCoordinates.

modules/commandModule/tests/testCommandModule.py
```diff
@@ -0,0 +1,44 @@
...
+        self.__pogiData['currentAirspeed'] = [round(random.uniform(0, 50), 7)]
+        self.__pogiData['eulerAnglesOfPlane'] = [random.uniform(0, math.pi * 2), random.uniform(0, math.pi * 2), random.uniform(0, math.pi * 2)]
+        self.__pogiData['eulerAnglesOfCamera'] = [random.uniform(0, math.pi * 2), random.uniform(0, math.pi * 2), random.uniform(0, math.pi * 2)]
+        self.__pogiData['isLanded'] = [random.randint(0,1)]
+
+    def write_to_json(self):
+        with open(self.pogiFileDirectory, 'w') as file:
+            json.dump(self.__pogiData, file, ensure_ascii=False, indent=4, sort_keys=True)
+
+mockData = mockCommandModule("./randomPogi.json")
```
FizzesShark, line 43: Do we want the file to be run to create a new random pogi file, or do we want the class to be used in other tests to create new random pogi files?

FizzesShark (changes requested):
Thanks for documenting what all the attributes are! But I do believe you're missing the currentAltitude field, and I think the eulerAnglesOfCamera should be a dictionary rather than an array, and the same with eulerAnglesOfPlane and gpsCoordinates.

modules/commandModule/tests/testCommandModule.py
```diff
@@ -0,0 +1,44 @@
...
+        list containing random airspeed in m/s
+    __pogiData['eulerAnglesOfPlane] : list
+        list containing, in order, random yaw, pitch, row angles of plane in radians
+    __pogiData['eulerAnglesOfCamera'] : list
+        list containing, in order, random yaw, pitch, row angles of camera in radians
+    __pogiData['isLanded'] : int
+        list containing either 1 for True or 0 for False to indicate whether drone has landed
+    """
+
+    def __init__(self, pogiFileDirectory: str):
```
Shrinjay, line 30: Ok, let's restructure this a little bit using the builder pattern. 

1) The constructor for this class should just set the pogiFileDirectory attribute. 
2) Need to break stuff up into functions. Define functions called get_gpsCoordinates, get_currentAirspeed, get_eulerAnglesofPlane, etc. for all the attributes of the pogi json. Each of these functions should return what is currently being assigned to each of these attributes. For example
`
get_gpsCoordinates():
   return {'latitude': random.uniform(-90, 90), 'longitude': random.uniform(-180, 180), 'altitude': random.uniform(0, 100)}
`
3) Define a function called build_pogi(). It should generate a dictionary and save it as a json using the ```write_to_json()``` function. self.__pogiData should no longer be a class attribute. It'll look something like:
```
build_json():
   pogi = {
      gpsCoordinates: get_gpsCoordinates(),
      currentAirspeed: get_currentAirspeed(),
      ....
   }
   write_to_json(pogi)
   return pogi
```

Shrinjay (changes requested):
Overall, really well done, you nailed the requirements in one pass. Just needs some minor restructuring and fixes. Tysm!

PR #83, opened 2021-09-03
Convert unittests to pytest

FizzesShark:
Changed file names such that pytest is able to discover tests; fixed multiple files that had various issues; deleted faulty test case that would've taken too much effort to fix.


sahil-kale (changes requested):
Can we add this to our CI workflow with Junit XML report generation?

Shrinjay:
> 
> 
> Can we add this to our CI workflow with Junit XML report generation?

I think we're going to de-prioritize further CI work for dev work which Gary is handling. I'll just merge this as-is for now so we have some test coverage.

PR #84, opened 2021-09-25
Added USBInterface class

K1VN:


modules/usbInterface/usbInterface.py
```diff
@@ -0,0 +1,36 @@
+import usb.core
+import usb.util
+
+class USBInterface:
```
Shrinjay, line 4: This is pretty perfect, I'm just going to ask for one extension. Firmware told me a day or so ago they want to do UART as well, so we'll implement both. Here's how I want to do it.

First, define a base class CommsInterface, with the same methods as USB Interface. Make them all abstract methods. Make USBInterface a subclass of CommsInterface
Then, implement the same logic for UART, creating endpoints, read and write according to the [PySerial Docs](https://pyserial.readthedocs.io/en/latest/pyserial.html#overview) and [the confluence doc](https://uwarg-docs.atlassian.net/wiki/spaces/CV/pages/1952022529/STM32-Jetson+Communication+Architecture)

modules/commsInterface/commsInterface.py
```diff
@@ -0,0 +1,74 @@
+import usb.core
+import usb.util
+import serial
```
sahil-kale, line 3: Please add a comment stating this is Pyserial, a requirements.txt file would be good as well so someone can just pip install the dependencies inside the text file

modules/commsInterface/commsInterface.py
```diff
@@ -0,0 +1,74 @@
...
+            return True
+
+class USBInterface(CommsInterface):
+    def __init__(self, type: bool, idVendor=None, idProduct=None):
+        self.idVendor = idVendor
+        self.idProduct = idProduct
+        super().__init__(self, type)
+
+class UARTInterface(CommsInterface):
+    def __init__(self, type: bool, uart_port, baudrate):
```
sahil-kale, line 69: Can you add a type hint for baudrate and keep it as a number?

modules/commsInterface/commsInterface.py
```diff
@@ -0,0 +1,74 @@
...
+    def read(self, endpointId):
+        """
+        Issues a write command in either serial or pyusb.
+        Data parameter should be a buffer. 
+        """
+        if self.uart_or_usb == 0:
+            return self.dev.read(endpointId, self.ep.wMaxPacketSize)
+        elif self.uart_or_usb == 1:
+            read_data = endpointId.readline()
+            return str(read_data[0:len(read_data)].decode('utf-8'))
```
sahil-kale, line 47: I don't think you should decode this data as a string. Reason being, Firmware and CV will come up with an API that is likely a raw byte stream (i.e. the number '232' is going to be '0xE8', which will decode to some random ASCII character).

sahil-kale (changes requested):
Few minor changes.

Can we get a document on confluence linked and start working on the API?

Additionally, one question - why are we passing in an endpointID rather than keeping it as a private object in the class? Not sure if this is a regular design pattern or not, but typically I'd expect to create a comms instance and not have to worry about keeping track of a specific object.

sahil-kale:
Also, you could probably nuke out the USB code since we are going with UART

sahil-kale:
> > Also, you could probably nuke out the USB code since we are going with UART
> 
> No I told him to keep it, just in case we switch our minds lmao

Copy that LOL

sahil-kale:
> I think we're just thinking of interfaces differently, I'm seeing this more as a library rather than a class, so it doesn't hold any state, so the endpoint isn't a state we hold either. It just holds a bunch of methods, and I just use the class because it's cleaner than having raw functions which you import. That means we could really go with either depending on usability.

Yup I think you summarized it perfectly, my first instinct was to treat it like a class rather than a lib since that's what I typically do. It is packed as as a USB interface class rather than a python module though...

 I'll leave final implementation authority to you and the team, I'm just getting my feet wet with the CV codebase lol

modules/commsInterface/commsInterface.py
Shrinjay: Good catch, but then we need to develop a lookup for all our mappings from raw bytes to values?

modules/commsInterface/commsInterface.py
```diff
@@ -0,0 +1,74 @@
...
+        if self.uart_or_usb == 0:
+            self.dev.write(endpointId, data)
+            return True
+        elif self.uart_or_usb == 1:
+            endpointId.write(data.encode())
+            endpointId.close()
+            return True
+
+class USBInterface(CommsInterface):
+    def __init__(self, type: bool, idVendor=None, idProduct=None):
```
Shrinjay, line 63: Why are we passing type in an as an argument? We know that if it's a USB interface, type is always 0 and if it's UART, type is always 1. Better to put these in literals like ``` super().__init__(self, 0) ``` for the USB interface constructor.

modules/commsInterface/commsInterface.py
```diff
@@ -0,0 +1,74 @@
...
+            0|False for USB, 1|True for UART
+        """
+        self.uart_or_usb = type
+    
+    def create_end_point_FC(self):
+        """
+        Issues a read command in either serial (UART) or pyusb (USB).
+        EndpointId can be the specific serial port to use or the usb endpoint to be used. 
+        """
+        if self.uart_or_usb == 0:
```
Shrinjay, line 23: Honestly this isn't quite how I was envisioning you'd do this but it's kinda cool so we'll keep it. As long as it works.

modules/commsInterface/commsInterface.py
sahil-kale: I don't understand what you mean by lookup? The way you'd typically do this is have a python struct definition on the computer side detailing what to expect in byte form. Something like
- 4x bytes unsigned integers, 1x float big endian etc
that is developed as a result of the FW/CV API. Then, you'd do struct.unpack() on the bytes object and store the results in various variables.

I guess this means there will be a master lookup somewhere, this is typically done with python structs though https://docs.python.org/3/library/struct.html

modules/commsInterface/commsInterface.py
sahil-kale: I could probably be ultra nitpicky and say use an enum, but I don't care enough cause we're using UART lol

PR #85, opened 2021-09-28
Created framePreProc module 

mikashaw:
I have no idea if my code makes any sense... but here it is


modules/framePreProc/framePreProc.py
```diff
@@ -0,0 +1,48 @@
...
+            dictionary holding the previous eulerAnglesOfCamera to compare to
+
+    """
+
+    def __init__(self, eulerDict, eulerDictLast):
+
+        self.eulerDict = eulerDict
+        self.eulerDictLast = eulerDictLast
+
+    def filter(self):
```
Shrinjay, line 24: Instead of storing self.eulerDict, pass in the current eulerDict as an argument.

modules/framePreProc/framePreProc.py
```diff
@@ -0,0 +1,48 @@
...
+        yaw = abs(self.eulerDictLast['yaw'] - self.eulerDict['yaw'])
+        pitch = abs(self.eulerDictLast['pitch'] - self.eulerDict['pitch'])
+        roll = abs(self.eulerDictLast['roll'] - self.eulerDict['roll'])
+            
+        if (yaw >= THRESHHOLD or pitch >= THRESHHOLD or roll >= THRESHHOLD):
+            return False
+            
+        else:
+            return True 
+
```
Shrinjay, line 45: Add a function called update_last_dict(eulerDict) that sets the self.lastEulerDict.

modules/framePreProc/framePreProcWorker.py
```diff
@@ -0,0 +1,35 @@
...
+
+    telemetryDataObject: MergedData object
+        Includes the telemetry data + euler angles data via .telemetry
+
+    pipelineOut: Queue?
+        Destination of telemetry data successfully passing through the filter
+
+    """
+
+    global lastTelemetry
```
Shrinjay, line 22: Here's how I'd restructure this. I wouldn't maintain lastTelemetry anywhere in the worker function. Instead, first get the eulerAnglesDict from the input data, then pass it into the filter function, then pass it into the update_last_dict function. If it passes the filter, put the input data in the output pipeline. It should look like this.

```
def frame_pre_proc_worker(pipelineIn, pipelineOut):
   frame_proc = FramePreProc()
   while true:
      dataObj = pipelineIn.get()
      eulerAnglesDict = dataObj.telemetry['eulerAnglesOfCamera']
      filter_pass = frame_proc.filter(eulerAnglesDict)
      frame_pre_proc.update_last_dict(eulerAnglesDict)
      
      if filter_pass:
         pipelineOut.put(dataObj)
      else:
         continue
```

modules/framePreProc/framePreProc.py
```diff
@@ -0,0 +1,48 @@
...
+        """
+
+        Finds the difference between present euler angles and those of the previous state.
+
+        Returns
+        -------
+        Boolean: 'True' if absolute difference per angle does not exceed 30 degrees else returns 'False'
+
+        """
+
```
Shrinjay, line 35: Add a check here, if self.lastEulerDict is None, then just return None as we can't perform processing.

Shrinjay (changes requested):
Most of the way there, just minor changes

PR #87, opened 2021-10-22
CI-Workflow-1

sahil-kale:


Shrinjay:
I decided to just get rid of the linter for now so we could check for build and test on merge easily. I'll also create the lint issue

PR #90, opened 2022-01-30
Add FW-CV Comms Code off UART interface

Shrinjay:
Note, we're just waiting for FW to confirm that the output from this code looks alright, then I'll update the test. We can merge this right away if we want, though.

### DIFF
--- a/modules/commsInterface/commsInterface.py
+++ b/modules/commsInterface/commsInterface.py
@@ -5,7 +5,7 @@ from digi.xbee.devices import XBeeDevice
 
 
 class CommsInterface:
-    def __init__(self, usb_type: bool):
+    def __init__(self, usb_type: int):
         """
         Initializes parameters for USB or UART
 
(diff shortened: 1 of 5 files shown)


tests/test_fltconn_send.py
```diff
@@ -0,0 +1,17 @@
...
+        "data": {
+            "longitude": 1.23,
+            "latitude": 1.23,
+            "qr_scan_flag": 1,
+            "takeoff_command": 0,
+            "detect_flag": 1
+        }
+    }
+
+    print(build_fijo_bytearray(command))
```
Shrinjay, line 16: We can update this test later to have a useful assertion, right now it just tests that the code doesn't crash

modules/communications/FltConnSend.py
```diff
@@ -0,0 +1,48 @@
...
+    detect_flag = request['data']['detect_flag']
+    takeoff_command = request['data']['takeoff_command']
+
+    def pack_to_int(val):
+        return struct.pack("i", val)
+
+    def pack_to_float(val):
+        return struct.pack("f", val)
+
+    byte_list = [
```
Shrinjay, line 38: Code that handles byte packing, incase you're from fw

modules/commsInterface/commsInterface.py
```diff
@@ -66,14 +66,14 @@ def __init__(self, idVendor=None, idProduct=None):
         # use lsusb -v in the terminal to list the idVendor and idProduct of connected usb devices 
         self.idVendor = idVendor
         self.idProduct = idProduct
-        super().__init__(self, 0)
+        super().__init__(0)
```
rayjinghaolei, line 69: not really a comment but isn't "self" needed for instanciating the subclass?

rayjinghaolei (commented):
not very familiar with the details of it but looks good overall

PR #93, opened 2022-02-19
Created searchExplosive module

krishna-kalavadia:
Created a module to detect objects using edge detection. Included sample images in which parcels are photoshopped (poorly) onto a drone view of grass.

Main steps are as follows:
1. Try to remove as much noise as possible so only objects are detected not patterns in the grass 
2. Perform Canny edge detection 
3. Perform contour detection and filter out any unwanted contours 
4. Draw bounding box on detected contours 

Example output:
Screen Shot 2022-02-21 at 4 27 30 PM


modules/searchExplosive/searchExplosive.py
```diff
@@ -0,0 +1,112 @@
+import cv2
+import numpy as np
+
```
rayjinghaolei, line 3: Consider making everything into a class and put those functions inside a class

rayjinghaolei (changes requested):
Also I tested all those four images, is there a way that you can remove those extremely small bounding boxes?
maybe we can do something like check the size of the bounding box compared to the entire frame size and if it's less than a certain threshold like 5% (you can change this number) of the width and height of the frame

modules/searchExplosive/searchExplosive.py
```diff
@@ -0,0 +1,118 @@
...
+            # Get rid of any unwanted contour detections if their bounding box is less than 0.1% of the image area
+            if (rectArea / imgArea) * 100 < 0.1:
+                continue
+
+            # Draw bounding box around the detected contours
+            cv2.rectangle(self.image, (x, y), (x + w, y + h), BLUE, 2)
+            cv2.putText(image, "Object", (x, y - 5), cv2.FONT_HERSHEY_SIMPLEX, 0.4, BLUE)
+
+        cv2.imshow('Window', self.image)  # show for testing
```
rayjinghaolei, line 108: sorry for the late review, but I think you can remove the imshow and put the contour into a class field like self.cnt or return it. Both is fine.

PR #94, opened 2022-03-05
First draft of ground-send code

Shrinjay:


modules/communications/groundSend.py
```diff
@@ -0,0 +1,48 @@
+from ..commsInterface.commsInterface import XBeeInterface
+import struct
+
+DEVICE_PORT = 0
```
rayjinghaolei, line 4: This needs to be filled in when testing right?

rayjinghaolei (commented):
Otherwise should be good

PR #96, opened 2022-03-08
First draft of DeepSort Integration

MichaelAiz:
Why: 
- We need to integrate DeepSort into our repo 

What: 
- Added Yolov5_DeepSort_Pytorch as submodule
- created new personDetection which closely follows the structure of Pylon detection
- replaced pylonDetection with personDetection inside of targetAcquisition 
- Added unit test for deepSORT tracker


mikashaw (changes requested):
Looks great! only thing would be to move the tests into  the general tests folder for computer-vision-python (we want to be able to run all the tests at once eventually). Other than that awesome job.

PR #97, opened 2022-03-11
framepreproc unit test

aydanjiwani:


tests/test_framepreproc.py
```diff
@@ -0,0 +1,31 @@
+import unittest
+import numpy as np
+
+from modules.framePreProc import framePreProc
+class TestFilter(unittest.TestCase):
```
mikashaw, line 5: let's call this 'TestFramePreProc'

tests/test_framepreproc.py
```diff
@@ -0,0 +1,31 @@
+import unittest
```
mikashaw, line 1: For WARG unit testing, we are trying to keep things consistent by using pytest, I just made a unit testing protocol doc, so just refer to https://uwarg-docs.atlassian.net/wiki/spaces/CV/pages/2036695089/Unit+Testing+Protocol
(basically, just import pytest and don't use unittest). Other than that, basically everything else looks good, so just make the minor changes and resumbit the PR!

mikashaw (changes requested):
refer to comments!

PR #104, opened 2022-03-24
Fixed calculation for centre of bounding box

MichaelAiz:
Why:
- Our previous formula for calculating the centre of a bounding box was not really finding the centre

What:
- Divided the previous result by 2 to find the actual centre of the bbox


MichaelAiz:
Also updated the comment for bounding box

PR #105, opened 2022-03-31
refactored get_non_collinear_points()

jeffreyli0312:
Modified files geolocation.py and test_geolocation.py, refactored the get_non_collinear_points() function in geolocation.py according to Ryan's Specification 2 in #geolocation. It now returns an array of indexes that we want to find to create a subarray of an (n, 2, 2) array in run_locator(). test_geolocation.py was changed accordingly to assert the output array of indexes, for ex [0, 1, 2, 3].

### DIFF
--- a/modules/geolocation/geolocation.py
+++ b/modules/geolocation/geolocation.py
@@ -214,6 +214,7 @@ class Geolocation:
         self.__logger.debug("geolocation/get_non_collinear_points: Started")
 
         NUM_POINTS_NEEDED = 4
+        indexes = []
 
         # If there aren't four points, return the empty array
         if len(coordinatesArray) < NUM_POINTS_NEEDED:
@@ -231,20 +232,38 @@ class Geolocation:
             for j in range(0, NUM_POINTS_NEEDED):
                 points[j] = coordinatesArray[(i + j) % len(coordinatesArray)]
 
-            # Check collinearity of all possible combinations
             areNotFourCollinear = True
-            for i in range(0, NUM_POINTS_NEEDED):
-                areNotFourCollinear &= not self.__are_three_points_collinear(points[i],
-                                                                             points[(i + 1) % NUM_POINTS_NEEDED],
-                                                                             points[(i + 2) % NUM_POINTS_NEEDED])
+            index0 = None
+            index1 = None
+            index2 = None
+            index3 = None
+
+            # Check collinearity of all possible combinations
+            for k in range(0, NUM_POINTS_NEEDED):
+                areNotFourCollinear &= not self.__are_three_points_collinear(points[k],
+                                                                             points[(k + 1) % NUM_POINTS_NEEDED],
+                                                                             points[(k + 2) % NUM_POINTS_NEEDED])
+                # Store indexes of current iteration
+                index0 = k
+                index1 = (k + 1) % NUM_POINTS_NEEDED
+                index2 = (k + 2) % NUM_POINTS_NEEDED
+                index3 = (k + 3) % NUM_POINTS_NEEDED
+
+                # If points are colinear, stop looping
+                if (not areNotFourCollinear):
+                    break
 
             # If all four points are non-collinear, return this combination of points
             if areNotFourCollinear:
                 self.__logger.debug("geolocation/get_non_collinear_points: Returned " + str(points))
-                return points
+                indexes = [index0, index1, index2, index3]
+
+                # Sort and return the indexes in ascending order
+                indexes.sort()
+                return indexes
         
         self.__logger.debug("geolocation/get_non_collinear_points: Returned np.empty(shape=(0,2))")
-        return np.empty(shape=(0, 2))
+        return indexes
 
     def calculate_pixel_to_geo_mapping(self):
         """
@@ -585,10 +604,24 @@ class Geolocation:
         if len(point_pairs) < 4:
             return False, None
 
-        non_collinear_points = self.get_non_collinear_points(point_pairs)
+        # Slice point_pairs to shape (n, 2) for input of get_non_collinear_points
+        points = point_pairs[:,1]
+
+        # Get the 4 non-collinear indexes of the array above
+        indexes = self.get_non_collinear_points(points)
+
         # If insufficient point pairs, exit this run and try again
-        if len(non_collinear_points) < 4:
+        if len(indexes) < 4:
             return False, None
+
+        non_collinear_points = np.empty(shape=(0, 2, 2))
+        # Create a subset of the (n, 2, 2) array above using the array of indexes
+        for i in len(indexes):
+            for j in len(point_pairs):
+                if i == j:
+                    non_collinear_points = np.concatenate((non_collinear_points, point_pairs[j]))
+                    # non_collinear_points only stores the 4 non-collinear point pairs
+                    # indicated by the indexes array
 
         self.__pixelToGeoPairs = non_collinear_points
         tranformation_matrix = self.calculate_pixel_to_geo_mapping()
--- a/modules/geolocation/geolocation.py
+++ b/modules/geolocation/geolocation.py
@@ -243,7 +243,7 @@ class Geolocation:
                 indexes[0] = k
                 indexes[1] = (k + 1) % NUM_POINTS_NEEDED
                 indexes[2] = (k + 2) % NUM_POINTS_NEEDED
-                indexes[3] = (k + 3) % NUM_POINTS_NEEDED #err0r?
+                indexes[3] = (k + 3) % NUM_POINTS_NEEDED
 
                 # If points are collinear, stop looping
                 if (areNotFourCollinear == False):
--- a/modules/geolocation/geolocation.py
+++ b/modules/geolocation/geolocation.py
@@ -619,7 +619,7 @@ class Geolocation:
         for i in range(0, len(indexes)):
             for j in range (0, len(point_pairs)):
                 if i == j:
-                    non_collinear_points = np.concatenate((non_collinear_points, point_pairs[j]))
+                    non_collinear_points = np.concatenate((non_collinear_points, [point_pairs[j]]))
                     # non_collinear_points only stores the 4 non-collinear point pairs
                     # indicated by the indexes array
 
--- a/modules/geolocation/geolocation.py
+++ b/modules/geolocation/geolocation.py
@@ -214,7 +214,7 @@ class Geolocation:
         self.__logger.debug("geolocation/get_non_collinear_points: Started")
 
         NUM_POINTS_NEEDED = 4
-        indexes = [0, 0, 0, 0] #this or np.empty?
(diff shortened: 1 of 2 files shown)


modules/geolocation/geolocation.py
```diff
@@ -231,20 +232,38 @@ def get_non_collinear_points(self, coordinatesArray):
             for j in range(0, NUM_POINTS_NEEDED):
                 points[j] = coordinatesArray[(i + j) % len(coordinatesArray)]
 
-            # Check collinearity of all possible combinations
             areNotFourCollinear = True
-            for i in range(0, NUM_POINTS_NEEDED):
-                areNotFourCollinear &= not self.__are_three_points_collinear(points[i],
-                                                                             points[(i + 1) % NUM_POINTS_NEEDED],
-                                                                             points[(i + 2) % NUM_POINTS_NEEDED])
+            index0 = None
```
Xierumeng, line 236: Why are there new variables when indexes can be used to store the current iteration? Remove this.

modules/geolocation/geolocation.py
```diff
@@ -231,20 +232,38 @@ def get_non_collinear_points(self, coordinatesArray):
...
+            index2 = None
+            index3 = None
+
+            # Check collinearity of all possible combinations
+            for k in range(0, NUM_POINTS_NEEDED):
+                areNotFourCollinear &= not self.__are_three_points_collinear(points[k],
+                                                                             points[(k + 1) % NUM_POINTS_NEEDED],
+                                                                             points[(k + 2) % NUM_POINTS_NEEDED])
+                # Store indexes of current iteration
+                index0 = k
```
Xierumeng, line 247: Store inside indexes instead of new variables.

modules/geolocation/geolocation.py
```diff
@@ -231,20 +232,38 @@ def get_non_collinear_points(self, coordinatesArray):
...
+
+                # If points are colinear, stop looping
+                if (not areNotFourCollinear):
+                    break
 
             # If all four points are non-collinear, return this combination of points
             if areNotFourCollinear:
                 self.__logger.debug("geolocation/get_non_collinear_points: Returned " + str(points))
-                return points
+                indexes = [index0, index1, index2, index3]
```
Xierumeng, line 259: Can be removed if stored in indexes instead of temporary variables.

modules/geolocation/geolocation.py
```diff
@@ -231,20 +232,38 @@ def get_non_collinear_points(self, coordinatesArray):
...
+
+                # Sort and return the indexes in ascending order
+                indexes.sort()
+                return indexes
         
         self.__logger.debug("geolocation/get_non_collinear_points: Returned np.empty(shape=(0,2))")
-        return np.empty(shape=(0, 2))
+        return indexes
```
Xierumeng, line 266: Replace with `return []` for default failure.

modules/geolocation/geolocation.py
```diff
@@ -585,11 +604,25 @@ def run_locator(self, telemetry, coordinates):
...
+        indexes = self.get_non_collinear_points(points)
+
         # If insufficient point pairs, exit this run and try again
-        if len(non_collinear_points) < 4:
+        if len(indexes) < 4:
             return False, None
 
+        non_collinear_points = np.empty(shape=(0, 2, 2))
+        # Create a subset of the (n, 2, 2) array above using the array of indexes
+        for i in range(0, len(indexes)):
```
Xierumeng, line 619: Replace loops with just `non_collinear_points = point_pairs[indexes]` to extract the points we want (see [this answer](https://stackoverflow.com/a/3179137)).

As a side note that won't be applied given the above:

`i == j` is a bug and should be `indexes[i] == j` . But if we already know what the indexes are we don't actually need the inner loop, and can just take them with `point_pairs[indexes[i]]` instead. Also, `numpy.append()` and `numpy.concatenate()` is expensive (underneath, a new array is allocated with a bigger size and elements copied over); since we already know the size it would be better to start with an empty array of size 4 instead of 0.

tests/test_geolocation.py
```diff
@@ -168,7 +168,8 @@ def test_four_points_collinear(self):
 
     def test_four_points_non_collinear(self):
         coordinatesArray = np.array([[0, 0], [10, 12], [-1, -1], [100, 0]])
-        expected = np.array([[0, 0], [10, 12], [-1, -1], [100, 0]])
+        # expected = np.array([[0, 0], [10, 12], [-1, -1], [100, 0]])
```
Xierumeng, line 171: Remove the line.

tests/test_geolocation.py
```diff
@@ -178,7 +179,8 @@ def test_four_points_non_collinear(self):
 
     def test_more_than_four_points_one_correct_case(self):
         coordinatesArray = np.array([[0, 0], [10, 12], [-1, -1], [100, 0], [4, 4]])
-        expected = np.array([[0, 0], [10, 12], [-1, -1], [100, 0]])
+        # expected = np.array([[0, 0], [10, 12], [-1, -1], [100, 0]])
```
Xierumeng, line 182: Remove the line.

I also would like you to add one more case: `test_more_than_four_points_last_correct_case` with:
* `coordinatesArray = np.array([[0, 0], [1, 1], [100, 0], [-1, -1], [4, 4], [2, -2], [-4, 10]])`
* `expected = np.array([0, 1, 2, 6])`

modules/geolocation/geolocation.py
```diff
@@ -214,6 +214,7 @@ def get_non_collinear_points(self, coordinatesArray):
         self.__logger.debug("geolocation/get_non_collinear_points: Started")
 
         NUM_POINTS_NEEDED = 4
+        indexes = [0, 0, 0, 0] #this or np.empty?
```
Xierumeng, line 217: Use `numpy.empty()` since the expected output is a numpy array.

modules/geolocation/geolocation.py
```diff
@@ -231,20 +232,33 @@ def get_non_collinear_points(self, coordinatesArray):
...
-                                                                             points[(i + 1) % NUM_POINTS_NEEDED],
-                                                                             points[(i + 2) % NUM_POINTS_NEEDED])
+
+            # Check collinearity of all possible combinations
+            for k in range(0, NUM_POINTS_NEEDED):
+                areNotFourCollinear &= not self.__are_three_points_collinear(points[k],
+                                                                             points[(k + 1) % NUM_POINTS_NEEDED],
+                                                                             points[(k + 2) % NUM_POINTS_NEEDED])
+                # Store indexes of current iteration
+                indexes[0] = k
```
Xierumeng, line 243: `i` instead.

modules/geolocation/geolocation.py
```diff
@@ -231,20 +232,33 @@ def get_non_collinear_points(self, coordinatesArray):
...
+                indexes[2] = (k + 2) % NUM_POINTS_NEEDED
+                indexes[3] = (k + 3) % NUM_POINTS_NEEDED
+
+                # If points are collinear, stop looping
+                if (areNotFourCollinear == False):
+                    break
 
             # If all four points are non-collinear, return this combination of points
-            if areNotFourCollinear:
+            if areNotFourCollinear == True:
```
Xierumeng, line 253: Unnecessary.

modules/geolocation/geolocation.py
```diff
@@ -231,20 +232,33 @@ def get_non_collinear_points(self, coordinatesArray):
...
                 self.__logger.debug("geolocation/get_non_collinear_points: Returned " + str(points))
-                return points
+
+                # Sort and return the indexes in ascending order
+                indexes.sort()
+                return indexes
         
         self.__logger.debug("geolocation/get_non_collinear_points: Returned np.empty(shape=(0,2))")
-        return np.empty(shape=(0, 2))
+        return []
```
Xierumeng, line 261: Return `np.empty(0)` .

modules/geolocation/geolocation.py
```diff
@@ -585,11 +599,23 @@ def run_locator(self, telemetry, coordinates):
...
+
+        # Get the 4 non-collinear indexes of the array above
+        indexes = self.get_non_collinear_points(points)
+
         # If insufficient point pairs, exit this run and try again
-        if len(non_collinear_points) < 4:
+        if len(indexes) < 4:
             return False, None
 
+        non_collinear_points = np.empty(shape=(4, 2, 2))
```
Xierumeng, line 612: Remove.

Xierumeng, line 612: Use numpy's array indexing: `np.array(indexes)` .

tests/test_geolocation.py
```diff
@@ -198,6 +198,16 @@ def test_more_than_four_points_no_correct_cases(self):
         # Test
         np.testing.assert_almost_equal(actual, expected)
 
+    def test_more_than_four_points_last_correct_case(self):
+        coordinatesArray = np.array([[0, 0], [1, 1], [100, 0], [-1, -1], [4, 4], [2, -2], [-4, 10]])
```
Xierumeng, line 202: Should be [-2, -2], whoops.

modules/geolocation/geolocation.py
```diff
@@ -585,11 +599,23 @@ def run_locator(self, telemetry, coordinates):
...
+        if len(indexes) < 4:
             return False, None
 
+        non_collinear_points = np.array(indexes)
+
+        # Create a subset of the (n, 2, 2) array above using the array of indexes
+        non_collinear_points = [point_pairs[i] for i in indexes]
+        # non_collinear_points only stores the 4 non-collinear point pairs
+        # indicated by the indexes array
+
```
Xierumeng, line 618: Sorry, what I meant is that you can take the subset from an array using the square bracket `[]` operator, except instead of of a number inside, you can put an array and it will take the elements.

Example:
```python
a = np.array([2, 4, 6, 8])
b = np.array([0, 3])
print(a[b]) # Outputs: [2 8]
```

modules/geolocation/geolocation.py
```diff
@@ -231,20 +232,33 @@ def get_non_collinear_points(self, coordinatesArray):
...
+            # Check collinearity of all possible combinations
+            for k in range(0, NUM_POINTS_NEEDED):
+                areNotFourCollinear &= not self.__are_three_points_collinear(points[k],
+                                                                             points[(k + 1) % NUM_POINTS_NEEDED],
+                                                                             points[(k + 2) % NUM_POINTS_NEEDED])
+                # Store indexes of current iteration
+                indexes[0] = (i)
+                indexes[1] = ((i + 1) % len(coordinatesArray))
+                indexes[2] = ((i + 2) % len(coordinatesArray))
+                indexes[3] = ((i + 3) % len(coordinatesArray))
```
Xierumeng, line 246: Move out of the k loop. Can be turned into its own loop (actually, you can make the j loop do double-duty).

Xierumeng:
Approved.

PR #110, opened 2022-04-13
Target acquisition to geolocation

jeffreyli0312:
Changed the write_locations() function to write to CSV file in format:
longitude, latitude
which is the required format in the KML file used to draw the mapLabelling lines. All contents of the CSV file will be copy pasted into the KML.

Added file CSV2KML.kml will be a file that contents are copied into, between the coordinates tag. The KML will be imported into google maps and the latitude/longitude will be drawn as a red line. Note that altitude in the KML file was found to be unnecessary, and so it was removed from both the CSV file and KML file.

Edit (update): All CSV files have been removed, as the write_locations() function now creates a KML file that is saved in the mapLabelling folder using the input coordinate array and simplekml package:
pip install simplekml

Note that latitude and longitude had to be switched in the lat_lon_from_local() function for the KML file to write in correct order. The KML file that is created can now be directly imported into google maps without any code changes.


modules/geolocation/geolocation.py
```diff
@@ -5,6 +5,8 @@
 import numpy as np
 import math
 import logging
+import os.path
```
Xierumeng, line 8: How necessary is this?

Additionally, don't forget to merge/rebase main into this branch.

modules/geolocation/geolocation.py
```diff
@@ -680,7 +682,15 @@ def map_location_from_pixel(self, transformationMatrix, pixels):
...
+                                  description='This is the path of the intruder.',
+                                  coords = locations) # Using list of long/lat
+
+        lines.style.linestyle.width = 3
+        lines.style.linestyle.color = simplekml.Color.red
+
+        kml.save(completeName) # Save KML file using completeName given
+
         return True
+
```
Xierumeng, line 696: Nitpick: Delete extra line, only one line at end of file.

Xierumeng:
Approved.

PR #114, opened 2022-06-01
updated venv instruction, readme and test typo

rayjinghaolei:


sahil-kale:
May I suggest that we make a bash script for setup that executes these commands automatically?

tests/targetAcquisition_geolocation_test.py
```diff
@@ -65,4 +65,4 @@ def test_targetAcquisition_to_geolocation(get_image):
     assert coordinates_and_telemetry != None
 
     assert check2 == True
-    assert geo_coordinates != None
```
Shrinjay, line 68: Are these tests all stable?

Shrinjay (commented):
Other thing, I thought we used pytorch now not tensorflow?

tests/targetAcquisition_geolocation_test.py
rayjinghaolei: I think on my laptop they are stable.
The reason that we still include tensorflow is that they are still used in our code. 
I installed them for it to pass the test

rayjinghaolei:
> May I suggest that we make a bash script for setup that executes these commands automatically?
I gave that a try, I don't think git bash works out well for the venv setup and it requires two scripts instead of one for windows and Mac/linux so I don't see much value doing it. Our environment setup is very short and manageable as well.
We are likely going to remove the git submodule thing as soon as possible, that reduced the work for setup even more.
Doing a setup is good step for new joiners, it helps to introduce them the idea of virtual environment and setting up an environment is always a necessary step in almost every project. It's a good place to let them start learning about things and increase more member-lead interaction.
There can be many other reasons, TLDR. but I don't really see a point right now to do this, we can do it in the future if setting up becomes complicated

PR #119, opened 2022-11-25
Style guide fixes after programming with Pylint

Xierumeng:
Having had to program with Pylint on, it's very annoying to have to add `pylint: disable=` and `pylint: enable=` everywhere. But you get used to it like `std::` and the code is so much cleaner.


aydanjiwani (approved):
LGTM

PR #125, opened 2023-04-18
YOLOv8 landing pad training

Xierumeng:


mgupta27 (commented):
Should we specify epochs to smth like 40 rather than default which is 100. Can we also specify batch size to be 32 rather than default which is 16. https://docs.ultralytics.com/modes/train/#usage-examples.

PR #128, opened 2023-06-14
Formatting and fix tests

Xierumeng:


tests/test_detect_target_worker.py
```diff
@@ -15,7 +15,7 @@
 # from modules import points_and_time
 
 
-MODEL_PATH = "tests/model_example/yolov8s.pt"
+MODEL_PATH = "tests/model_example/yolov8s_pretrained_default.pt"
```
mgupta27, line 18: Do you plan on creating a PR with the landing pad model?

mgupta27 (approved):
LGTM. Thank you for doing the formatting. :)

tests/test_detect_target_worker.py
Xierumeng: No, there are multiple models and they're pretty large. I was considering not including this one either but then the tests won't run properly without some extra setup.

PR #129, opened 2023-06-14
Made predictions device agnostic

mgupta27:
Made predictions device agnostic so that the code runs on machines with and without CUDA. After setting up the code on my laptop I ran into this error so I fixed it so no one should run into these errors now.

### DIFF
--- a/modules/detect_target/detect_target.py
+++ b/modules/detect_target/detect_target.py
@@ -5,6 +5,7 @@ import time
 
 import cv2
 import numpy as np  # TODO: Remove
+import torch
 import ultralytics
 
 from .. import frame_and_time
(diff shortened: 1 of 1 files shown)


modules/detect_target/detect_target.py
```diff
@@ -22,6 +23,7 @@ def __init__(self, model_path: str, save_name: str=""):
         self.__model = ultralytics.YOLO(model_path)
         self.__counter = 0
         self.__filename_prefix = save_name + "_" + str(int(time.time())) + "_"
+        self.__device = 0 if torch.cuda.is_available() else "cpu"
```
Xierumeng, line 26: Nit: Move to the line after `self.__model`

Personally, I don't like this format of conditional assignment but that's just my preference, fine to leave as is.

Xierumeng:
Please rebase on or merge from latest main to resolve conflicts.

Xierumeng:
Next time also please use the WARG repository directly rather than a personal fork.

modules/detect_target/detect_target.py
mgupta27: Ok in that case we can leave it. Maybe we should just make a note of it in the documentation so people know.

Xierumeng (approved):

PR #130, opened 2023-06-24
Interprocess communication messages

Xierumeng:
Structs passed between workers.

### DIFF
--- a/modules/detection_in_world.py
+++ b/modules/detection_in_world.py
@@ -1,0 +1,28 @@
+"""
+Object detection in world space
+"""
+
+import numpy as np
+
+
+# Basically a struct
+# pylint: disable=too-few-public-methods
+class DetectionInWorld:
+    """
+    Typically on the ground
+    """
+    def __init__(self, vertices: np.ndarray, centre: np.ndarray, label: int, confidence: float):
+        """
+        vertices is a quadrilateral of 4 points
+        centre is a point
+        A point is an xy coordinate (index 0 and 1 respectively)
+        """
+        assert vertices.shape == (4, 2)
+        assert centre.shape[0] == 2
+
+        self.vertices = vertices
+        self.centre = centre
+        self.label = label
+        self.confidence = confidence
+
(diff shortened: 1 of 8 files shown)


modules/odometry_and_time.py
```diff
@@ -0,0 +1,61 @@
...
+    Specifically, intrinsic (Tait-Bryan) rotations in the zyx/3-2-1 order
+    """
+    def __init__(self, yaw: float, pitch: float, roll: float):
+        # TODO: Verify bounds
+        assert yaw >= -180.0
+        assert yaw <= 180.0
+        assert pitch >= -180.0
+        assert pitch <= 180.0
+        assert roll >= -180.0
+        assert roll <= 180.0
```
Xierumeng, line 37: DO NOT MERGE UNTIL RESOLVED.

Xierumeng: Verified, MAVLink message ID 30 (Attitude).

modules/detection_in_world.py
```diff
@@ -0,0 +1,28 @@
...
+    Typically on the ground
+    """
+    def __init__(self, vertices: np.ndarray, centre: np.ndarray, label: int, confidence: float):
+        """
+        vertices is a quadrilateral of 4 points
+        centre is a point
+        A point is an xy coordinate (index 0 and 1 respectively)
+        """
+        assert vertices.shape == (4, 2)
+        assert centre.shape[0] == 2
```
mgupta27, line 21: Just a concern about using `assert` in the code base. If in any case the assertion is false an `AssertionError` will be raised and the airside code will stop running. Is it possible to maybe put a `try catch block` in the codebase to handle the assertions errors when they pop up? That way our code can handle any problems it encounters and still run.

Xierumeng: We want to avoid try-catch blocks as much as possible for performance reasons. Additionally, it is preferable to invalidate bad data as soon as possible.

mgupta27 (approved):
LGTM.

PR #131, opened 2023-06-25
CUDA check

Xierumeng:
CUDA available assertion to ensure that computer has CUDA.


mgupta27 (approved):
LGTM. Nit: Can we add this in the confluence document so that people know about this step.

mgupta27 (approved):
LGTM

PR #135, opened 2023-07-05
Worker exit on sentinel

Xierumeng:
If a worker encounters the sentinel value (`None` ) from the queue, exit immediately.


mgupta27 (approved):
LGTM!

PR #141, opened 2023-07-21
Geolocation refactor

Xierumeng:


modules/geolocation/camera_properties.py
```diff
@@ -0,0 +1,257 @@
...
+            pixel_y,
+            self.resolution_y,
+            self.__vec_v,
+        )
+        if not result:
+            return False, None
+
+        # Get Pylance to stop complaining
+        assert vec_pixel_u is not None
+        assert vec_pixel_v is not None
```
mgupta27, line 195: Is it possible that when we check if `vec_pixel_u` or `vec_pixel_v` is not None we return `False, None` instead of using an assertion. I bring this up because if the assertion fails then we stop the airside system in flight so it would be nice if we can handle the error instead.

modules/geolocation/geolocation.py
```diff
@@ -0,0 +1,330 @@
...
+        output_vertices = (perspective_transform_matrix @ input_vertices).T
+
+        ground_centre = np.array(
+            [
+                output_centre[0] / output_centre[2],
+                output_centre[1] / output_centre[2],
+            ],
+            dtype=np.float32,
+        )
+
```
mgupta27, line 222: This is the part of the code I don't really understand. Can you please explain why we have to divide by `output_centre[2]`

modules/geolocation/geolocation.py
```diff
@@ -0,0 +1,330 @@
...
+                [top_left[0], top_left[1], 1.0],
+                [top_right[0], top_right[1], 1.0],
+                [bottom_left[0], bottom_left[1], 1.0],
+                [bottom_right[0], bottom_right[1], 1.0],
+            ],
+            dtype=np.float32,
+        ).T
+
+        # Single row/column does not need transpose
+        output_centre = perspective_transform_matrix @ input_centre
```
mgupta27, line 211: Not a huge issue but could we change the spelling from `centre` to `center`

mgupta27 (approved):
LGTM! Just left a few comments on some questions I had about the code.

modules/geolocation/camera_properties.py
Xierumeng: Yes, these assertions should never fail as they are guarded by line 190 above. They are to get the IDE linter to stop complaining about the variables possibly having a type of None and therefore being unusable.

modules/geolocation/geolocation.py
```diff
@@ -0,0 +1,330 @@
...
+        ):
+            # Last row must be 0, 0, 1 for matrix to be affine
+            return False, None
+
+        centre = detection.get_centre()
+        top_left, top_right, bottom_left, bottom_right = detection.get_corners()
+
+        input_centre = np.array([centre[0], centre[1], 1.0], dtype=np.float32)
+        # More efficient to multiply a matrix than looping over the points
+        # Transpose to columns from rowss
```
Xierumeng, line 199: Nit: Spelling.

modules/geolocation/geolocation.py
Xierumeng: `centre` is the current Canadian spelling.

modules/geolocation/geolocation.py
Xierumeng: This is a homogeneous/perspective divide (see: https://en.wikipedia.org/wiki/Transformation_matrix#Perspective_projection ). I can add a comment.

modules/geolocation/geolocation.py
mgupta27: Oh, you're right ... I've been using the American spelling

PR #144, opened 2023-08-20
Formatting cleanup

Xierumeng:


modules/image_and_time.py
```diff
@@ -0,0 +1,40 @@
...
+import numpy as np
+
+
+# Basically a struct
+# pylint: disable=too-few-public-methods
+class ImageAndTime:
+    """
+    Contains image and timestamp.
+    """
+    __create_key = object()
```
mgupta27, line 15: Do we want to write a test file for this class?

mgupta27: Actually never mind. There really isn't much to test as this class is mostly used for holding information.

mgupta27 (approved):

modules/image_and_time.py
Xierumeng: Yes, but out of scope of this PR.

PR #160, opened 2024-01-29
Fix worker controller exit request deadlock

Xierumeng:
Multiple requests to exit without clearing causes deadlock. This change fixes this.


mgupta27 (approved):
LGTM!

PR #164, opened 2024-02-18
Fix data merge worker test

Xierumeng:
New merge class is not created if the provided detections class lacks detections. Simulated detect target worker now creates a detection.


mgupta27 (approved):
LGTM!

PR #167, opened 2024-02-26
Detect target precision profile

KarthiU:
### Created profiler_detect_target_2024.py and modified detect_target.py to add timing functionality. 
**profiler_detect_target_2024.py**
- added dependencies on timing, pandas, and numpy libraries
- Structured on around the main_2024.py file, but removed flight interface related code
- Created a profiler.txt file that holds raw data used for analysis 
- Added indicator to config.yaml to stop the main loop after x seconds have passed (define in the config) 
- Gets logged data and interprets it, outputting the min, max, average, median and initial times for preprocessing, inference, postprocessing and total elapsed time of the detect_target prediction loop 

**detect_target.py**
- If full precision is not overriden to true, it defaults to false (for testing purposes only) 
- saves timing data to profiler.txt 

**config.yaml**
- added profiling_time to set how long you profile for (i.e. 300 = 5 min)


modules/detect_target/detect_target.py
```diff
@@ -75,6 +80,22 @@ def run(self, data: image_and_time.ImageAndTime) -> "tuple[bool, np.ndarray | No
                 assert detection is not None
                 detections.append(detection)
 
+        stop_time = time.time()
+
+        elapsed_time = stop_time - start_time
+      
+        for pred in predictions: 
+            with open('profiler.txt', 'a') as file:
+                speeds = pred.speed
```
mgupta27, line 89: We don't want to add logic within the `detect_target.py` class for profiling. Rather we could time the worker outside of the call to detect target.

modules/detect_target/detect_target.py
```diff
@@ -27,9 +27,12 @@ def __init__(self, device: "str | int", model_path: str, override_full: bool, sa
         self.__device = device
         self.__model = ultralytics.YOLO(model_path)
         self.__counter = 0
-        self.__enable_half_precision = False if self.__device == "cpu" else True
+        self.__enable_half_precision = False if self.__device == "cpu" else False
+        #modified so override_full controls if its half or full - FOR PROFILING ONLY
```
mgupta27, line 31: Instead of modifying the logic here, when profiling why not just specify the device type and the current logic will handle if half or full precision should be used?

profiler_detect_target_2024.py
```diff
@@ -0,0 +1,203 @@
...
+
+
+CONFIG_FILE_PATH = pathlib.Path("config.yaml")
+
+
+def main() -> int:
+    """
+    copied from airside code main function
+    """
+    # Open config file
```
mgupta27, line 29: Do you think you could explain what the difference between this file is and `profile_detect_target.py`? Just trying to understand what the difference is.

modules/detect_target/detect_target.py
Xierumeng: It might be worth updating this interface to a `create()` method that takes a settings object as argument. The settings class could have a `is_setting_valid()` method, which can push the responsibility of ensuring a valid setting to the caller. This also increases transparency because there won't be any fallthrough logic (the setting is applied as is).

modules/detect_target/detect_target.py
Xierumeng: The worker isn't the thing to test, something like this:
```py
# profiling_or_whatever_file_name.py

images: "list[np.ndarray]" = load_300_images()
detector = detect_target.DetectTarget(...)
for image in images:
    gc.disable()  # This disables the garbage collector
    start = time.time_ns()
    result, value = detector.run(image)  # Might or might not want to keep the bounding boxes
    end = time.time_ns()
    gc.enable()  # This enables the garbage collector
    if not result:
        # Handle error
  
    # Save results somewhere
    time_ns = end - start
    ...
```

Xierumeng:
It would also be better to use a dataset that are not all identical images. Perhaps the same as used in repository profile-encode.

PR #168, opened 2024-03-01
Search pattern

nsmartinx:


modules/decision/search_pattern.py
```diff
@@ -0,0 +1,97 @@
+from .. import decision_command
```
mgupta27, line 1: Can you add a docstring here to explain what this class does.

modules/decision/search_pattern.py
```diff
@@ -0,0 +1,97 @@
...
+        # If it is at it's current target location, update to the next target
+        if self.distance_to_target_squared(current_position) < self.acceptable_variance_squared:
+            self.set_target_location()
+
+        # Send command to go to target
+        return decision_command.DecisionCommand.create_move_to_absolute_position_command(
+            self.target_posx,
+            self.target_posy,
+            current_position.odometry_data.position.down)
+
```
mgupta27, line 97: Nit: Add a line at the end of the file.

modules/decision/search_pattern.py
```diff
@@ -0,0 +1,97 @@
...
+        # Calculate x and y coordinates of new target location
+        self.relative_target_posx = self.search_radius * cos(self.angle)
+        self.relative_target_posy = self.search_radius * sin(self.angle)
+
+        self.target_posx = self.search_origin.north + self.relative_target_posx
+        self.target_posy = self.search_origin.east + self.relative_target_posy
+    
+    def distance_to_target_squared(self,
+                                   current_position: odometry_and_time.OdometryAndTime
```
mgupta27, line 73: Nit: Add a comma at the end here

mgupta27 (commented):
Reviewed.

modules/decision/search_pattern.py
```diff
@@ -0,0 +1,98 @@
...
+from .. import decision_command
+from .. import odometry_and_time
+from math import tan, pi, cos, sin, ceil
+
+class SearchPattern:
+    """
+    Attributes:
+        camera_fov (float): 
+            Camera's field of view, measured in degrees. Use the smallest measurement available.
+            Is essentially assumed to be a circle
```
TongguangZhang, line 14: does the field of view refer to the the angle between the line to the furthest point the camera can see and straight down, or double that angle (I'm assuming it's double the angle)? You don't have to change anything here, just remember to put this in the documentation

modules/decision/search_pattern.py
```diff
@@ -0,0 +1,98 @@
...
+                 acceptable_variance_squared: float):
+        
+        # Initialize the drone to the first position in the search pattern
+        self.current_ring = 0
+        self.current_pos_in_ring = 0
+        self.max_pos_in_ring = 0
+        self.search_radius = 0
+        self.acceptable_variance_squared = acceptable_variance_squared
+        
+        # Store the origin of the search and
```
TongguangZhang, line 40: Nit: comments should be complete sentences (with periods at the end)

modules/decision/search_pattern.py
```diff
@@ -0,0 +1,98 @@
...
+        Updates the target position to the next position in the search pattern.
+        """
+
+        # If it is done the current ring, move to the next ring. If not, next step in current ring
+        if self.current_pos_in_ring >= self.max_pos_in_ring:
+            self.current_ring += 1
+            self.current_pos_in_ring = 0
+            self.search_radius = self.search_gap * self.current_ring
+            self.max_pos_in_ring = (ceil(self.search_radius * 2 * pi / self.search_gap))
+            self.angle_between_positions = ((2 * pi) / (self.max_pos_in_ring + 1))
```
TongguangZhang, line 60: remove extra brackets in these lines

TongguangZhang, line 60: Can you walk me through the math you did here? since the search pattern isn't actually a circle but an inscribed polygon, the circumference `self.search_radius * 2 * pi` would be bigger than the perimeter of the polygon right? This gives us too many search points - we're going in more than a circle.

modules/decision/search_pattern.py
```diff
@@ -0,0 +1,98 @@
...
+
+        # Calculate x and y coordinates of new target location
+        self.relative_target_posx = self.search_radius * cos(self.angle)
+        self.relative_target_posy = self.search_radius * sin(self.angle)
+
+        self.target_posx = self.search_origin.north + self.relative_target_posx
+        self.target_posy = self.search_origin.east + self.relative_target_posy
+    
+    def distance_to_target_squared(self,
+                                   current_position: odometry_and_time.OdometryAndTime
```
TongguangZhang, line 75: Make this a private static method since it's a helper function and isn't changing the class - update the name to be __distance_to_target_squared (double underscore indicates private method)

modules/decision/search_pattern.py
```diff
@@ -0,0 +1,98 @@
...
+        acceptable_variance_squared (float):
+            The square of the acceptable distance from the target position.
+    """
+
+    def __init__(self,
+                 camera_fov: float,
+                 search_height: float,
+                 search_overlap: float,
+                 current_position: odometry_and_time.OdometryAndTime,
+                 acceptable_variance_squared: float):
```
TongguangZhang, line 31: variance refers more to a probability/statistics concept, could you rename this to something like distance_square_threshold?

tests/test_search_pattern.py
```diff
@@ -0,0 +1,99 @@
...
+        assert search_maker.max_pos_in_ring == 0
+
+    def test_distance_to_target_squared(self, search_maker, drone_odometry):
+        """
+        Test distance_to_target_squared method.
+        """
+        # Drone and target position are at the origin, so the distance should be near 0
+        
+        distance = search_maker.distance_to_target_squared(drone_odometry)
+        assert distance <= 0.1
```
TongguangZhang, line 69: set up tests in the format

1. setup - get the objects you need and set the expected value, hard code the expected value with expected = True
2. run the functions to get the actual value - distance = search_maker ..., and set actual = distance <= 0.1
3. assertions - assert that actual == expected

this is just an example for this test case, do this for the other test cases too

modules/decision/search_pattern.py
```diff
@@ -0,0 +1,193 @@
...
+        self.moving_horizontally = self.current_side_in_square % 2 == 0
+
+        # Calculate the length of the current side based on the direction of movement
+        if self.moving_horizontally:
+            side_length = next_corner[0] - self.current_corner[0]
+        else:
+            side_length = next_corner[1] - self.current_corner[1]
+
+        # Calculate the number of stops needed along the current side
+        self.max_pos_on_side = ceil(abs(side_length) / self.search_gap_depth)
```
TongguangZhang, line 118: the number of stops should just be the multiple right?

TongguangZhang (commented):
Reviewed

modules/decision/search_pattern.py
```diff
@@ -0,0 +1,242 @@
...
+
+        # If the depth is less than the width, we apply an offset at each corner to ensure the
+        # entire area is scanned
+        adjustment = 0
+        if self.__search_gap_depth < self.__search_width:
+            adjustment = (self.__search_gap_width - self.__search_gap_depth) / 2
+
+        # Calculate the corners based on the offsets and the search origin. Top left corner is moved
+        # right by search_gap_width as the final side of the square will instead cover that part
+        self.__square_corners = [
```
TongguangZhang, line 108: a few options here to make it clear which is the top left, top right, bottom left, bottom right:

1. use an object instead that stores the current state (tl, tr, bl, br) with enums
2. use a dict with an enum class for the keys
3. have these as class variables with a state variable that's assigned enum values

by enums I mean have an object with constants for the possible states so instead of assigning state = "top left", you would have state = enum.TOP_LEFT

modules/decision/search_pattern.py
```diff
@@ -0,0 +1,242 @@
...
+    def __calculate_side_of_square(self):
+        """
+        Computes the gaps along the current side of the square
+        """
+        # Calculate the positions of the current and next corners
+        self.__current_corner = self.__square_corners[self.__current_side_in_square]
+        next_corner = self.__square_corners[(self.__current_side_in_square + 1) % 4]
+
+        # Determine if the drone is moving horizontally or vertically along the current side
+        self.__moving_horizontally = self.__current_side_in_square % 2 == 0
```
TongguangZhang, line 139: from above - having it hardcoded would make this more readable so we can tell which direction (clockwise vs counter clockwise) we're going

modules/decision/search_pattern.py
```diff
@@ -0,0 +1,242 @@
...
+        # Calculate the corners based on the offsets and the search origin. Top left corner is moved
+        # right by search_gap_width as the final side of the square will instead cover that part
+        self.__square_corners = [
+            (
+                self.__search_origin_x
+                - square_size
+                - adjustment
+                + self.__search_gap_width,
+                self.__search_origin_y + square_size,
+            ),  # Top left corner
```
TongguangZhang, line 115: use an object instead of tuples to clarify which is. x and which is y (I know it's pretty conventional to have (x, y) but using indexes does not tell the reader what the indexed values mean)

modules/decision/search_pattern.py
```diff
@@ -0,0 +1,242 @@
...
+            side_length = next_corner[0] - self.__current_corner[0]
+        else:
+            side_length = next_corner[1] - self.__current_corner[1]
+
+        # As calculated, it is the distance to the next corner, however, we want to stop before
+        # we actually reach the next side as the area will be covered by it
+        if side_length > 0:
+            side_length += (self.__search_gap_depth - self.__search_gap_width) / 2
+        else:
+            side_length -= (self.__search_gap_depth - self.__search_gap_width) / 2
```
TongguangZhang, line 153: why do we subtract here, also, if this is the very first case and is never used again, we don't want to evaluate the if every time, so let's cover it in the init (maybe by setting it to this value in the first place

modules/decision/search_pattern.py
```diff
@@ -0,0 +1,242 @@
...
+            # Increment the position counter
+            self.__current_pos_on_side += 1
+            return False
+
+        # Calculate the next target position based on the current fraction of the side covered
+        dist_to_move = self.__travel_gap * self.__current_pos_on_side
+        if self.__moving_horizontally:
+            self.__target_posx = self.__current_corner[0] + dist_to_move
+        else:
+            self.__target_posy = self.__current_corner[1] + dist_to_move
```
TongguangZhang, line 209: I would prefer having using x += self.__travel_gap  and y += self.__travel_gap since these should represent the current state of the drone to avoid the multiplications and reassigning x and y

tests/test_search_pattern.py
```diff
@@ -0,0 +1,172 @@
...
+    """
+    search_pattern_instance = search_pattern.SearchPattern(
+        camera_fov_forwards=30.0,
+        camera_fov_sideways=60.0,
+        search_height=100.0,
+        search_overlap=0.2,
+        current_position_x=100.0,
+        current_position_y=50.0,
+        distance_squared_threshold=2.0,
+        small_adjustment=1,
```
TongguangZhang, line 26: if you use these numbers to initialize every instance of the class, use a constant

tests/test_search_pattern.py
```diff
@@ -0,0 +1,172 @@
...
+        search_overlap=0.2,
+        current_position_x=75.0,
+        current_position_y=150.0,
+        distance_squared_threshold=2.0,
+        small_adjustment=1,
+    )
+    yield search_pattern_instance
+
+
+def create_drone_odometry(pos_x, pos_y, height):
```
TongguangZhang, line 49: use a fixture for this

tests/test_search_pattern.py
```diff
@@ -0,0 +1,172 @@
...
+import pytest
+
+from modules import decision_command
+from modules import odometry_and_time
+from modules import drone_odometry_local
+from modules.decision import search_pattern
+
+
+@pytest.fixture()
+def search_pattern_width_greater_height():
```
TongguangZhang, line 14: if you call it depth in the class, keep the naming consistent

TongguangZhang (commented):
reviewed

modules/decision/search_pattern.py
nsmartinx: This is because the way I'm currently doing it the side length is signed (i.e. positive when moving up/right or negative when moving down/left).

tests/test_search_pattern.py
```diff
@@ -0,0 +1,187 @@
...
+def assert_move_command(move, newPos, target_x, target_y, target_z):
+    assert move[0] == newPos
+    assert (
+        move[1].get_command_type()
+        == decision_command.DecisionCommand.CommandType.MOVE_TO_ABSOLUTE_POSITION
+    )
+    pos = move[1].get_command_position()
+    assert pytest.approx(target_x, 0.1) == pos[0]
+    assert pytest.approx(target_y, 0.1) == pos[1]
+    assert pytest.approx(target_z, 0.1) == pos[2]
```
TongguangZhang, line 83: set this up with the testing convention
1. inputs and expected - expected_move = newPos, etc.
2. actual values - actual_pos = move[0]
3. assert actual == expected

tests/test_search_pattern.py
```diff
@@ -0,0 +1,187 @@
...
+    Tests for the SearchPattern class methods
+    """
+
+    def test_width_greater_depth_normal(
+        self, search_pattern_width_greater_depth, drone_odometry
+    ):
+        """
+        Test first 20 positions of search pattern where drone has reached point before being called
+        """
+        coordinates = [
```
TongguangZhang, line 97: use expected_coordinates as the name here to follow testing convention and make it clear it is the expected output

tests/test_search_pattern.py
```diff
@@ -0,0 +1,187 @@
...
+            [32.38, 185.25, True],
+            [68.00, 185.25, True],
+            [103.62, 185.25, True],
+            [139.25, 185.25, True],
+            [174.87, 185.25, True],
+        ]
+
+        for i in range(len(coordinates) - 1):
+            odometry = drone_odometry(coordinates[i][0], coordinates[i][1], 100)
+            move = search_pattern_width_greater_depth.continue_search(odometry)
```
TongguangZhang, line 123: use actual_coordinates or another descriptive name here to make it clear this is the output

TongguangZhang (commented):
reviewed!

TongguangZhang (approved):
LGTM!

PR #169, opened 2024-03-24
Added multiprocessing logging module

TongguangZhang:
Added logging example in main_2024 and video_input

### DIFF
--- a/modules/multiprocess_logging/multiprocess_logging.py
+++ b/modules/multiprocess_logging/multiprocess_logging.py
@@ -7,12 +7,14 @@ WARNING = logging.WARNING
 ERROR = logging.ERROR
 CRITICAL = logging.CRITICAL
 
-def message_and_metadata(message, frame):
-        function_name = frame.f_code.co_name
-        filename = frame.f_code.co_filename
-        line_number = inspect.getframeinfo(frame).lineno
 
-        return f'[{filename} | {function_name} | {line_number}] {message}'
+def message_and_metadata(message, frame):
+    function_name = frame.f_code.co_name
+    filename = frame.f_code.co_filename
+    line_number = inspect.getframeinfo(frame).lineno
+
+    return f"[{filename} | {function_name} | {line_number}] {message}"
+
 
 def log_message(message, level, frame, queue):
-        queue.queue.put((message_and_metadata(message, frame), level), block=False)
+    queue.queue.put((message_and_metadata(message, frame), level), block=False)
--- a/modules/multiprocess_logging/multiprocess_logging.py
+++ b/modules/multiprocess_logging/multiprocess_logging.py
@@ -1,0 +1,9 @@
+import inspect
+import linecache
+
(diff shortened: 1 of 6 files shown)


modules/multiprocess_logging/multiprocess_logging.py
```diff
@@ -0,0 +1,18 @@
...
+WARNING = logging.WARNING
+ERROR = logging.ERROR
+CRITICAL = logging.CRITICAL
+
+def message_and_metadata(message, frame):
+        function_name = frame.f_code.co_name
+        filename = frame.f_code.co_filename
+        line_number = inspect.getframeinfo(frame).lineno
+
+        return f'[{filename} | {function_name} | {line_number}] {message}'
```
TongguangZhang, line 15: the frame object cannot be pickled so you can't put it in a queue, this is why we need this helper function, since we probably don't want to write this every time we want to log

modules/multiprocess_logging/multiprocess_logging.py
```diff
@@ -0,0 +1,18 @@
+import inspect
+import logging
+
+DEBUG = logging.DEBUG
+INFO = logging.INFO
+WARNING = logging.WARNING
+ERROR = logging.ERROR
+CRITICAL = logging.CRITICAL
```
TongguangZhang, line 8: this so so that modules trying to log don't have to import the logging module just for the logging levels

modules/multiprocess_logging/multiprocess_logging.py
TongguangZhang, line 15: the frame object cannot be pickled so you can't put it in a queue, this is why we need this helper function, since we probably don't want to write this every time we want to log

modules/video_input/video_input.py
```diff
@@ -1,24 +1,41 @@
...
     def run(self) -> "tuple[bool, image_and_time.ImageAndTime | None]":
         """
         Returns a possible ImageAndTime with current timestamp.
         """
         result, image = self.device.get_image()
+
+        try:
+            frame = inspect.currentframe()
+            multiprocess_logging.log_message(f'image size {image.shape}', multiprocess_logging.DEBUG, frame, self.logging_queue)
+            # self.logging_queue.queue.put((multiprocess_logging.message_and_metadata(f'image size {image.shape}', frame), logging.DEBUG), block=False)
```
TongguangZhang, line 35: these are the 2 methods I'm thinking about using right now - the second method is more consistent with our current modules, where we put things in queues, the first method is more similar to how the default logging module is used

TongguangZhang (commented):
I can't really decide how to implement this custom logging module - should I try to make it as similar to the default logging module as possible or make it consistent with our other modules?

modules/multiprocess_logging/multiprocess_logging.py
mgupta27: Nit: Can we create a separate file and store these values as an enum?

PR #170, opened 2024-03-29
Logging module

TongguangZhang:


modules/logger/logger.py
```diff
@@ -0,0 +1,135 @@
...
+
+        # Get the path to the logs directory.
+        log_directory_path = config["log_directory_path"]
+        entries = os.listdir(log_directory_path)
+        log_names = [
+            entry for entry in entries if os.path.isdir(os.path.join(log_directory_path, entry))
+        ]
+
+        # Find the log directory for the current run, which is the most recent timestamp
+        datetime_format = "%Y-%m-%d_%H-%M-%S"
```
mgupta27, line 53: Nit: How about we make this a global constant since it is a hard-coded string.

modules/logger/logger.py
```diff
@@ -0,0 +1,135 @@
...
+        log_path = max(
+            log_names,
+            key=lambda datetime_string: datetime.datetime.strptime(
+                datetime_string, datetime_format
+            ),
+        )
+
+        filename = f"{log_directory_path}/{log_path}/{name}.log"
+
+        # Formatting configurations for the logger
```
mgupta27, line 63: Nit: I see some comments with periods and some that don't have periods. Can we try to make all the comments consistent?

mgupta27 (commented):
Reviewed.

main_2024.py
```diff
@@ -3,6 +3,8 @@
 """
 
 import argparse
+import datetime
+import inspect
```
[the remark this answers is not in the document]
TongguangZhang: both of them are built-in python modules so we don't need to install them

mgupta27 (approved):
LGTM

PR #171, opened 2024-04-03
Integrate geolocation worker into main

DylanFinlay:
Changes made to main_2024.py to integrate the geolocation worker. This integration needs to be ground tested and then flight tested.

### DIFF
--- a/main_2024.py
+++ b/main_2024.py
@@ -213,29 +213,6 @@ def main() -> int:
 
     while True:
         try:
-            merged_data = data_merge_to_geolocation_queue.queue.get_nowait()
-        except queue.Empty:
-            merged_data = None
-
-        if merged_data is not None:
-            position = merged_data.odometry_local.position
-            orientation = merged_data.odometry_local.orientation.orientation
-            detections = merged_data.detections
-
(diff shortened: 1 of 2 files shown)


main_2024.py
```diff
@@ -159,7 +176,38 @@ def main() -> int:
...
+            GEOLOCATION_CAMERA_POSITION_X,
+            GEOLOCATION_CAMERA_POSITION_Y,
+            GEOLOCATION_CAMERA_POSITION_Z,
+        ),
+        (
+            GEOLOCATION_CAMERA_ORIENTATION_YAW,
+            GEOLOCATION_CAMERA_ORIENTATION_PITCH,
+            GEOLOCATION_CAMERA_ORIENTATION_ROLL,
+        ),
+    )
```
TongguangZhang, line 201: These create methods return a tuple - `True, object` if successful and `False, None` if unsuccessful.
Add a check for the result flag and return an error if the result is False

TongguangZhang (commented):
Reviewed

TongguangZhang (approved):
LGTM!

PR #173, opened 2024-05-27
Update test_landing_pad_tracking.py

deltag0:


tests/unit/test_landing_pad_tracking.py
```diff
@@ -124,9 +124,12 @@ def test_is_similar_positive_equal_to_threshold(self) -> None:
         them is equal to the distance threshold.
         """
         obj_1 = object_in_world.ObjectInWorld.create(0, 0, 0)[1]
+        assert obj_1 is not None
```
TongguangZhang, line 127: we want to have it like this:
result, obj_1 = ...
assert result

TongguangZhang (commented):
reviewed! did you find any other cases where create methods aren't being used correctly?

tests/unit/test_landing_pad_tracking.py
deltag0: Nope, that's all I found

PR #177, opened 2024-06-23
DRAFT: Tested Geolocation

mgupta27:
Test Geolocation Draft PR


DanielJPuratich:
Geolocation.__MIN_DOWN_COS_ANGLE should be higher? In wind or fast travel the drone has a good chance of exceeding 6 degree angle.

DylanFinlay:
Unless I'm tripping I don't think the model "tests/model_example/best-2s.pt" is actually in this repo. I'm assuming you just have it there locally but maybe we should just add it in?

PR #178, opened 2024-06-23
Graceful handling and notification of errors

AshishA26:
main_2024.py has been updated to recover from a worker crashing. main checks if all workers are alive in each managers. If a worker is dead, log the error and restart the worker (terminate and join the worker, fill and drain the related queues, and create and start a new worker).

### DIFF
--- a/main_2024.py
+++ b/main_2024.py
@@ -289,7 +289,8 @@ def main() -> int:
 
                 if manager == ManagerType.VIDEO_INPUT.value:
                     main_logger.error(
-                        "Video Input Worker is dead, attempting to restart.", frame
+                        "Video Input Worker is dead, attempting to restart.", 
+                        frame,
                     )
                     video_input_to_detect_target_queue.fill_and_drain_queue()
                     del video_input_manager
@@ -309,7 +310,8 @@ def main() -> int:
 
                 elif manager == ManagerType.DETECT_TARGET.value:
                     main_logger.error(
-                        "Detect Target Worker is dead, attempting to restart.", frame
+                        "Detect Target Worker is dead, attempting to restart.", 
+                        frame,
                     )
                     video_input_to_detect_target_queue.fill_and_drain_queue()
                     detect_target_to_data_merge_queue.fill_and_drain_queue()
@@ -333,7 +335,8 @@ def main() -> int:
 
                 elif manager == ManagerType.FLIGHT_INTERFACE.value:
                     main_logger.error(
-                        "Flight Interface Worker is dead, attempting to restart.", frame
+                        "Flight Interface Worker is dead, attempting to restart.", 
+                        frame,
                     )
                     flight_interface_to_data_merge_queue.fill_and_drain_queue()
                     del flight_interface_manager
@@ -353,7 +356,8 @@ def main() -> int:
 
                 elif manager == ManagerType.DATA_MERGE.value:
                     main_logger.error(
-                        "Data Merge Worker is dead, attempting to restart.", frame
+                        "Data Merge Worker is dead, attempting to restart.", 
+                        frame,
                     )
                     detect_target_to_data_merge_queue.fill_and_drain_queue()
                     flight_interface_to_data_merge_queue.fill_and_drain_queue()
@@ -375,7 +379,8 @@ def main() -> int:
 
                 elif manager == ManagerType.GEOLOCATION.value:
                     main_logger.error(
-                        "Geolocation Worker is dead, attempting to restart.", frame
+                        "Geolocation Worker is dead, attempting to restart.", 
+                        frame,
                     )
                     data_merge_to_geolocation_queue.fill_and_drain_queue()
                     geolocation_to_main_queue.fill_and_drain_queue()
--- a/main_2024.py
+++ b/main_2024.py
@@ -8,6 +8,7 @@ import inspect
 import multiprocessing as mp
 import pathlib
 import queue
+from enum import Enum
 
 import cv2
 import yaml
@@ -25,7 +26,6 @@ from modules.logger import logger
 from utilities.workers import queue_proxy_wrapper
 from utilities.workers import worker_controller
 from utilities.workers import worker_manager
-from enum import Enum
 
 CONFIG_FILE_PATH = pathlib.Path("config.yaml")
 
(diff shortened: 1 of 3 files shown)


main_2024.py
```diff
@@ -8,6 +8,7 @@
 import multiprocessing as mp
 import pathlib
 import queue
+from enum import Enum
```
TongguangZhang, line 11: take a look at https://uwarg-docs.atlassian.net/wiki/spaces/CV/pages/2253226033/Python+Style+Convention#Imports for import style convention (make sure imports are in alphabetical order within their section)

main_2024.py
```diff
@@ -241,6 +241,27 @@ def main() -> int:
...
+    class ManagerType(Enum):
+        """
+        Enum class for mapping index in managers_array to worker manager names.
+        """
+
+        VIDEO_INPUT = 0
+        DETECT_TARGET = 1
+        FLIGHT_INTERFACE = 2
+        DATA_MERGE = 3
+        GEOLOCATION = 4
```
TongguangZhang, line 253: place this enum class outside of the main function

main_2024.py
```diff
@@ -255,6 +276,130 @@ def main() -> int:
...
+            if not managers_array[manager].are_workers_alive():
+                managers_array[manager].terminate_workers()
+                managers_array[manager].join_workers()
+                managers_array[manager] = None
+
+                if manager == ManagerType.VIDEO_INPUT.value:
+                    main_logger.error(
+                        "Video Input Worker is dead, attempting to restart.",
+                        frame,
+                    )
```
TongguangZhang, line 293: make sure to follow the docs for the logging module, we need to get a new frame each time since the frame is information about the step currently running, and we don't want an outdated frame https://uwarg-docs.atlassian.net/wiki/spaces/CV/pages/2515959830/Logging+Module

main_2024.py
```diff
@@ -255,6 +276,130 @@ def main() -> int:
                 print("geolocation confidence: " + str(detection_world.confidence))
                 print("")
 
+        # Check if all workers are alive in each manager.
+        # If a worker is dead, log the error and restart the worker.
+        # Steps: Terminate and join the worker, fill and drain the
+        # related queues, and create and start a new worker.
+        for manager in range(number_of_managers):
```
TongguangZhang, line 283: you should create a helper function for this - on an error, pass in the needed info to restart the worker

main_2024.py
```diff
@@ -255,6 +276,130 @@ def main() -> int:
...
+            if not managers_array[manager].are_workers_alive():
+                managers_array[manager].terminate_workers()
+                managers_array[manager].join_workers()
+                managers_array[manager] = None
```
TongguangZhang, line 287: what if the worker restart fails? this would give an error if any of the value sin manager_array are None, so we should add a check for that too (in that case, we should also log that the restart failed)

main_2024.py
```diff
@@ -255,6 +276,130 @@ def main() -> int:
...
+                        1,
+                        video_input_worker.video_input_worker,
+                        (
+                            VIDEO_INPUT_CAMERA_NAME,
+                            VIDEO_INPUT_WORKER_PERIOD,
+                            VIDEO_INPUT_SAVE_PREFIX,
+                            video_input_to_detect_target_queue,
+                            controller,
+                        ),
+                    )
```
TongguangZhang, line 308: we should change the rest of main to use a variable for the args so it's reusable (and so we can pass the args into a function), so we would have a variable named video_input_worker_args, etc.

main_2024.py
```diff
@@ -255,6 +276,130 @@ def main() -> int:
...
+                managers_array[manager].terminate_workers()
+                managers_array[manager].join_workers()
+                managers_array[manager] = None
+
+                if manager == ManagerType.VIDEO_INPUT.value:
+                    main_logger.error(
+                        "Video Input Worker is dead, attempting to restart.",
+                        frame,
+                    )
+                    video_input_to_detect_target_queue.fill_and_drain_queue()
```
TongguangZhang, line 294: I don't think we want to fill the queue before draining, filling the queue with sentinel (None) values will take down the worker reading from that queue, so drain is ok, we also don't necessarily have to drain the queue since a worker that died wouldn't have put bad data into the queue, and draining the queue also blocks the other workers after it until the dead worker is back up

I'm open to suggestions on this one though, since I'm not 100% sure on how to handle queues of dead workers

TongguangZhang (commented):
reviewed

main_2024.py
AshishA26: I agree that we don't need to fill the queue before draining. In terms of draining, maybe we should just be draining the preceding queue to ensure that the preceding queue data wasn't what caused the worker to fail. As for draining the succeeding queues, I don't think it is necessary because (as you mentioned) a worker that died wouldn't have put bad data into the queue.

main_2024.py
AshishA26: Would you be able to clarify what I should create the helper function for? Would you like me to create a helper function for 1 worker, for each worker, or for the for loop?

main_2024.py
AshishA26: Would you like me to check if the worker restart failed, or if any of the elements in the array are None? Also, if it does fail to restart, should I let it go through the whole "while True" loop again or should I add a boolean flag that says whether to exit a sub loop or not?

TongguangZhang: oh this is my bad, a failed restart wouldn't put a None value in the array, the list of arrays would just be empty, and the heartbeat function would actually return True, which is not what we want

utilities/workers/worker_manager.py
```diff
@@ -48,3 +48,21 @@ def join_workers(self) -> None:
...
+
+    def are_workers_alive(self) -> bool:
+        """
+        Check if workers are alive.
+
+        Return: True if all workers are alive. False is any 1 worker is not alive.
+        """
+        for worker in self.__workers:
+            if not worker.is_alive():
+                return False
```
TongguangZhang, line 60: check if the list is empty as well - if the list is empty, it would also indicate that the thread is dead

main_2024.py
TongguangZhang: there are some common elements to restarting a dead worker

1. we log a message
2. we drain its input queue(s) (since the workers have difference numbers of queues, pass an array of queues to drain)
3. delete the manager
4. create a new manager and add it to the array of managers
5. create workers for the new manager

I was thinking depending on the ManagerType, we would pass different parameters into a function that does this, maybe smth like terminate_and_restart_worker()

You could also have another helper function that does the for loop with the if blocks and calls terminate_and_restart_worker, but that seems unnecessary and you would have to pass it basically everything which is unideal

TongguangZhang (commented):
reviewed

main_2024.py
TongguangZhang: that makes sense, I agree

main_2024.py
AshishA26: So, just making sure, I shouldn't change anything here, right?

PR #181, opened 2024-07-02
Add logging to flight interface

mgupta27:
## Add logging to flight interface

**Key Changes**
* Initialized logger in flight interface to log information
* Logged the initialization of the logger and the home location of the drone

**Why are these changes needed**
* Needed for debugging purposes, especially when we flight test geolocation

### DIFF
--- a/modules/flight_interface/flight_interface.py
+++ b/modules/flight_interface/flight_interface.py
@@ -52,10 +52,12 @@ class FlightInterface:
         # Get Pylance to stop complaining
         assert home_location is not None
 
-        frame = inspect.currentframe() 
+        frame = inspect.currentframe()
         flight_interface_logger.info(home_location, frame)
 
-        return True, FlightInterface(cls.__create_key, controller, home_location, flight_interface_logger)
+        return True, FlightInterface(
+            cls.__create_key, controller, home_location, flight_interface_logger
+        )
 
     def __init__(
         self,
--- a/modules/flight_interface/flight_interface.py
+++ b/modules/flight_interface/flight_interface.py
@@ -61,7 +61,7 @@ class FlightInterface:
         class_private_create_key: object,
         controller: flight_controller.FlightController,
         home_location: drone_odometry.DronePosition,
-        logger: logger.Logger,
+        flight_interface_logger: logger.Logger,
     ) -> None:
         """
         Private constructor, use create() method.
@@ -70,7 +70,7 @@ class FlightInterface:
 
         self.controller = controller
         self.__home_location = home_location
-        self.__logger = logger
+        self.__logger = flight_interface_logger
 
         frame = inspect.currentframe()
         self.__logger.info(self.__home_location, frame)
--- a/modules/flight_interface/flight_interface.py
+++ b/modules/flight_interface/flight_interface.py
@@ -2,10 +2,13 @@
 Creates flight controller and combines odometry data and timestamp.
 """
 
+import inspect
+
 from . import local_global_conversion
 from .. import odometry_and_time
 from ..common.mavlink.modules import drone_odometry
 from ..common.mavlink.modules import flight_controller
+from ..logger import logger 
 
 
 class FlightInterface:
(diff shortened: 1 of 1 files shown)


modules/flight_interface/flight_interface.py
```diff
@@ -35,6 +38,13 @@ def create(cls, address: str, timeout_home: float) -> "tuple[bool, FlightInterfa
         # Get Pylance to stop complaining
         assert home_location is not None
 
+        result, flight_interface_logger = logger.Logger.create("flight_interface")
+
+        if result:
+            frame = inspect.currentframe()
+            flight_interface_logger.info("flight interface logger initialized", frame)
+            flight_interface_logger.info(home_location, frame)
```
TongguangZhang, line 46: we should probably pass the logger and result into the init in case we want to use the logger throughout the class (we need to pass the result as well because we need to check the result before logging every time)

TongguangZhang, line 46: I think this should be 1 message, since it's 1 action (and 1 frame)

TongguangZhang (commented):
reviewed

modules/flight_interface/flight_interface.py
```diff
@@ -35,6 +38,13 @@ def create(cls, address: str, timeout_home: float) -> "tuple[bool, FlightInterfa
         # Get Pylance to stop complaining
         assert home_location is not None
 
+        result, flight_interface_logger = logger.Logger.create("flight_interface")
```
DanielJPuratich, line 41: If logger is created before the TCP port could we log issues with opening the port as well? In the past this has been an issue where we encountered errors eh?

modules/flight_interface/flight_interface.py
mgupta27: Good point. I was reading thru ur documentation and read that we should just create it in `__init__` so we don't block the process. Is this still true. Link to docs: https://uwarg-docs.atlassian.net/wiki/spaces/CV/pages/2515959830/Logging+Module.

mgupta27: Fixed.

modules/flight_interface/flight_interface.py
mgupta27: Good point. Created it first so that we can log the other errors in the create method.

modules/flight_interface/flight_interface.py
mgupta27: After thinking about it, I believe its better to create the logger in `create` and exit if it is `None`. If we create if in `__init__` and its None that means anytime we want to log something we have to do null check which can be easily missed.

Also, if we put it in `create` and don't exit if the logger is `None`, then we would pass in a `logger | None` into init which means that we would still have to do the null check again. Let me know if you agree or have other thoughts.

modules/flight_interface/flight_interface.py
```diff
@@ -21,27 +24,44 @@ def create(cls, address: str, timeout_home: float) -> "tuple[bool, FlightInterfa
         address: TCP address or port.
         timeout_home: Timeout for home location in seconds.
         """
+        result, flight_interface_logger = logger.Logger.create("flight_interface")
+        if not result:
+            return False, None
```
TongguangZhang, line 29: I think the idea is we don't want to return False, None if the log creation fails, since it's not a critical part of the system, we don't want it taking down the system

modules/flight_interface/flight_interface.py
```diff
@@ -21,27 +24,44 @@ def create(cls, address: str, timeout_home: float) -> "tuple[bool, FlightInterfa
         address: TCP address or port.
         timeout_home: Timeout for home location in seconds.
         """
+        result, flight_interface_logger = logger.Logger.create("flight_interface")
```
TongguangZhang, line 27: we would call this smth like create_logger_result, and pass it into the init

modules/flight_interface/flight_interface.py
```diff
@@ -21,27 +24,44 @@ def create(cls, address: str, timeout_home: float) -> "tuple[bool, FlightInterfa
...
+            flight_interface_logger.error("controller could not be created", frame)
             return False, None
 
         # Get Pylance to stop complaining
         assert controller is not None
 
         result, home_location = controller.get_home_location(timeout_home)
         if not result:
+            frame = inspect.currentframe()
+            flight_interface_logger.error("home_location could not be created", frame)
```
TongguangZhang, line 49: all log calls would go
```
if self.__create_logger_result:
    frame = inspect.currentframe()
    flight_interface_logger.error("home_location could not be created", frame)
```

TongguangZhang (commented):
reviewed

TongguangZhang (approved):
lgtm

PR #188, opened 2024-07-14
Fix incorrect logging directory for perception

Xierumeng:
Video input and detect target now receiving correct prefix for logging.


DylanFinlay (approved):
LGTM

PR #189, opened 2024-07-16
Add flight interface decision queue

Trotyl15:


main_2024.py
```diff
@@ -212,6 +216,9 @@ def main() -> int:
             FLIGHT_INTERFACE_TIMEOUT,
             FLIGHT_INTERFACE_BAUD_RATE,
             FLIGHT_INTERFACE_WORKER_PERIOD,
+            flight_interface_to_data_merge_queue,
+            flight_interface_decision_queue,
+            controller,
```
TongguangZhang, line 221: flight_interface_to_data_merge_queue should go in the output_queues arg, and flight_interface_decision_queue should go in the input_queues arg

modules/flight_interface/flight_interface.py
```diff
@@ -9,6 +9,7 @@
 from ..logger import logger
 from ..common.mavlink.modules import drone_odometry
 from ..common.mavlink.modules import flight_controller
+from ..decision_command import DecisionCommand
```
TongguangZhang, line 12: to follow the design pattern, use from .. import decision_command instead (to match the importing of odometry_and_time

modules/flight_interface/flight_interface.py
```diff
@@ -18,6 +19,12 @@ class FlightInterface:
 
     __create_key = object()
 
+    MOVE_TO_RELATIVE_POSITION = 0
+    MOVE_TO_ABSOLUTE_POSITION = 1
+    LAND_AT_CURRENT_POSITION = 2
+    LAND_AT_RELATIVE_POSITION = 3
+    LAND_AT_ABSOLUTE_POSITION = 4
```
TongguangZhang, line 26: make this an enum class

TongguangZhang (commented):
reviewed

modules/flight_interface/flight_interface.py
```diff
@@ -83,8 +88,14 @@ def run(self) -> "tuple[bool, odometry_and_time.OdometryAndTime | None]":
...
+
         assert odometry_local is not None
 
         return odometry_and_time.OdometryAndTime.create(odometry_local)
+
+    def apply_decision(self, decision: decision_command.DecisionCommand) -> None:
+        """
+        Apply the given decision command to the flight controller.
+        """
+        self.controller.upload_commands(decision)
```
TongguangZhang, line 101: I think flight controller expects a list of dronekit.Command objects, not a DecisionCommand object, couple options here:
1. we could either change the function in common to accept a list of DecisionCommand objects and use enums to determine what dronekit.Command object to send (I like this option)
2. change this function to determine what dronekit.Command object to send based on enums
3. change the interface between flight interface and decision to pass dronekit.Command objects instead of DecisionCommand objects

TongguangZhang (commented):
we should add unit tests for the apply_decision function

PR #191, opened 2024-07-20
Graceful handling and notification of errors

AshishA26:
worker_manager.py has been updated to recover from a worker crashing. Code checks if all workers are alive in each manager. If a worker is dead, log the error and restart the worker (terminate and join the worker, drain the input queues, and create and start a new worker).

### DIFF
--- a/log.txt
+++ b/log.txt
@@ -1,0 +1,3 @@
+[WinError 10061] No connection could be made because the target machine actively refused it sleeping
+[WinError 10061] No connection could be made because the target machine actively refused it sleeping
+Cannot connect to drone! Make sure the address/port is correct.
--- a/log.txt
+++ b/log.txt
@@ -1,3 +1,0 @@
-[WinError 10061] No connection could be made because the target machine actively refused it sleeping
-[WinError 10061] No connection could be made because the target machine actively refused it sleeping
-Cannot connect to drone! Make sure the address/port is correct.
--- a/utilities/workers/worker_manager.py
+++ b/utilities/workers/worker_manager.py
@@ -233,6 +233,8 @@ class WorkerManager:
                 # Terminate and join worker
                 worker.terminate()
                 worker.join()
+                del worker
+                worker = None
 
                 # Drain the preceding queues
                 input_queues = self.__worker_properties.get_input_queues()
--- a/utilities/workers/worker_manager.py
+++ b/utilities/workers/worker_manager.py
@@ -250,4 +250,3 @@ class WorkerManager:
                 if not result:
                     frame = inspect.currentframe()
                     self.__local_logger.error("Failed to restart " + worker_name_string, frame)
-                    return False, None
--- a/utilities/workers/worker_manager.py
+++ b/utilities/workers/worker_manager.py
@@ -226,11 +226,11 @@ class WorkerManager:
                 new_workers.append(worker)
                 continue
 
-            # Log the error
+            # Log dead worker
             frame = inspect.currentframe()
-            target_and_worker_name = self.__worker_properties.get_target_name() + " " + worker.name
+            target_and_worker_name = f"{self.__worker_properties.get_target_name()} {worker.name}"
             self.__local_logger.warning(
-                "Worker died, restarting " + target_and_worker_name,
+                f"Worker died, restarting {target_and_worker_name}",
                 frame,
             )
 
@@ -242,7 +242,7 @@ class WorkerManager:
             )
             if not result:
                 frame = inspect.currentframe()
-                self.__local_logger.error("Failed to restart " + target_and_worker_name, frame)
+                self.__local_logger.error(f"Failed to restart {target_and_worker_name}", frame)
                 return False
 
             # Append the new worker
--- a/utilities/workers/worker_manager.py
+++ b/utilities/workers/worker_manager.py
@@ -27,7 +27,6 @@ class WorkerProperties:
         output_queues: "list[queue_proxy_wrapper.QueueProxyWrapper]",
         controller: worker_controller.WorkerController,
         local_logger: logger.Logger,
-        worker_name: str,
     ) -> "tuple[bool, WorkerProperties | None]":
         """
         Creates worker properties.
@@ -39,7 +38,6 @@ class WorkerProperties:
         output_queues: Output queues.
         controller: Worker controller.
         local_logger: Existing logger from process.
-        worker_name: Name of the worker type.
 
         Returns the WorkerProperties object.
         """
@@ -59,7 +57,6 @@ class WorkerProperties:
             input_queues,
             output_queues,
             controller,
-            worker_name,
         )
 
     def __init__(
@@ -71,7 +68,6 @@ class WorkerProperties:
         input_queues: "list[queue_proxy_wrapper.QueueProxyWrapper]",
         output_queues: "list[queue_proxy_wrapper.QueueProxyWrapper]",
         controller: worker_controller.WorkerController,
-        worker_name: str,
     ) -> None:
         """
         Private constructor, use create() method.
@@ -84,7 +80,6 @@ class WorkerProperties:
         self.__input_queues = input_queues
         self.__output_queues = output_queues
         self.__controller = controller
-        self.__worker_name = worker_name
 
     def get_worker_arguments(self) -> "tuple":
         """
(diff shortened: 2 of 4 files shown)


utilities/workers/worker_manager.py
```diff
@@ -106,6 +111,18 @@ def get_worker_target(self) -> "(...) -> object":  # type: ignore
...
+        """
+        Returns the input queues.
+        """
+        return self.__input_queues
+
+    def get_worker_name(self) -> str:
+        """
+        Returns the name of the worker type.
+        """
+        return self.__worker_name
```
Xierumeng, line 124: Is it possible to get the name of the target function instead?

utilities/workers/worker_manager.py
```diff
@@ -195,3 +218,35 @@ def join_workers(self) -> None:
...
+                # Terminate and join worker
+                worker.terminate()
+                worker.join()
+                del worker
+                worker = None
+
+                # Drain the preceding queues
+                input_queues = self.__worker_properties.get_input_queues()
+                for queue in input_queues:
+                    queue.drain_queue()
```
Xierumeng, line 242: It is inefficient to drain the queues every time the loop encounters a dead worker. Pull this out of the loop so that it only needs to be done once.

Xierumeng (commented):
Reviewed.

utilities/workers/worker_manager.py
AshishA26: We would have to create getter in each worker I think. Would this be a good idea, I'm not sure?

Xierumeng: I tried this and it worked:
```py
def test():
    pass


if __name__ == "__main__":
    a = test
    print(a.__name__)
```
So I think you could have:
```py
def get_target_name(self) -> str:
    return self.__target.__name__
```

Xierumeng: Actually, you can directly use the worker's process name instead:
```py
for worker in self.__workers:
    print(worker.name)
```
Source: https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Process.name

utilities/workers/worker_manager.py
```diff
@@ -195,3 +218,35 @@ def join_workers(self) -> None:
...
+                # Log the error
+                frame = inspect.currentframe()
+                worker_name_string = self.__worker_properties.get_worker_name() + " worker"
+                self.__local_logger.error("Worker died, restarting " + worker_name_string, frame)
+
+                # Terminate and join worker
+                worker.terminate()
+                worker.join()
+                del worker
+                worker = None
```
Xierumeng, line 237: This seems sketchy, considering that failing to create a worker below leaves this in the list. I think it's better to copy the alive workers into a new list, and then create and append whatever number of workers is required to bring it back to the original size.

utilities/workers/worker_manager.py
```diff
@@ -195,3 +218,35 @@ def join_workers(self) -> None:
...
+
+                # Create a new worker
+                result, worker = WorkerManager.__create_single_worker(
+                    self.__worker_properties.get_worker_target(),
+                    self.__worker_properties.get_worker_arguments(),
+                    self.__local_logger,
+                )
+                if not result:
+                    frame = inspect.currentframe()
+                    self.__local_logger.error("Failed to restart " + worker_name_string, frame)
```
Xierumeng, line 247: This should return something so that the caller knows that there is a catastrophic failure.

utilities/workers/worker_manager.py
```diff
@@ -195,3 +218,35 @@ def join_workers(self) -> None:
...
+
+        # Drain the preceding queues
+        input_queues = self.__worker_properties.get_input_queues()
+        for queue in input_queues:
+            queue.drain_queue()
```
Xierumeng, line 252: Hmm, why is this necessary?

Xierumeng (commented):
Reviewed.

utilities/workers/worker_manager.py
AshishA26: Draining the preceding queue ensures that the preceding queue data wasn't what caused the worker to fail. As for draining the succeeding queues, I don't think it is necessary because a worker that died wouldn't have put bad data into the queue.

utilities/workers/worker_manager.py
AshishA26: using worker.name would print out "Worker died, restarting Process-5". I'm not sure how useful this would be, should we be concatenating worker.name with the target name?

Xierumeng: That would work.

log.txt
Xierumeng, line 1: Remove.

utilities/workers/worker_manager.py
```diff
@@ -219,34 +219,43 @@ def join_workers(self) -> None:
...
+        Returns whether the dead workers were able to be restarted.
         """
+        new_workers = []
         for worker in self.__workers:
-            if not worker.is_alive():
-                # Log the error
+            if worker.is_alive():
+                new_workers.append(worker)
+
+        del self.__workers
```
Xierumeng, line 233: I don't think this is necessary.

utilities/workers/worker_manager.py
```diff
@@ -195,3 +218,44 @@ def join_workers(self) -> None:
...
+        """
+        new_workers = []
+        for worker in self.__workers:
+            if worker.is_alive():
+                new_workers.append(worker)
+
+        del self.__workers
+        self.__workers = new_workers
+
+        while len(self.__workers) < self.__worker_properties.get_worker_count():
```
Xierumeng, line 236: Use a for loop instead.

Xierumeng (commented):
Reviewed.

utilities/workers/worker_manager.py
Xierumeng: Add a comment.

utilities/workers/worker_manager.py
```diff
@@ -195,3 +213,49 @@ def join_workers(self) -> None:
...
+        """
+        Check and restart dead workers.
+
+        Returns whether the dead workers were able to be restarted.
+        """
+        new_workers = []
+        for worker in self.__workers:
+            if worker.is_alive():
+                new_workers.append(worker)
+            else:
```
Xierumeng, line 227: Nit: Instead of else block:
```py
if worker.is_alive():
    new_workers.append(worker)
    continue

...
```

utilities/workers/worker_manager.py
```diff
@@ -195,3 +213,49 @@ def join_workers(self) -> None:
...
+                # Log the error
+                frame = inspect.currentframe()
+                target_and_worker_name = (
+                    self.__worker_properties.get_target_name() + " " + worker.name
+                )
+                self.__local_logger.error(
```
Xierumeng, line 233: I think warning or info level is more appropriate, because error level implies almost unrecoverable failure.

Xierumeng (commented):
Reviewed.

utilities/workers/worker_manager.py
```diff
@@ -195,3 +213,48 @@ def join_workers(self) -> None:
...
+
+        Returns whether the dead workers were able to be restarted.
+        """
+        new_workers = []
+        for worker in self.__workers:
+            if worker.is_alive():
+                new_workers.append(worker)
+                continue
+
+            # Log the error
```
Xierumeng, line 229: Update this comment. For example: `# Log dead worker`

utilities/workers/worker_manager.py
```diff
@@ -195,3 +213,48 @@ def join_workers(self) -> None:
...
+            frame = inspect.currentframe()
+            target_and_worker_name = self.__worker_properties.get_target_name() + " " + worker.name
+            self.__local_logger.warning(
+                "Worker died, restarting " + target_and_worker_name,
```
Xierumeng, line 233: Use f strings instead of string concatenation.

utilities/workers/worker_manager.py
```diff
@@ -195,3 +213,48 @@ def join_workers(self) -> None:
...
+
+            # Create a new worker
+            result, new_worker = WorkerManager.__create_single_worker(
+                self.__worker_properties.get_worker_target(),
+                self.__worker_properties.get_worker_arguments(),
+                self.__local_logger,
+            )
+            if not result:
+                frame = inspect.currentframe()
+                self.__local_logger.error("Failed to restart " + target_and_worker_name, frame)
```
Xierumeng, line 245: f strings.

Xierumeng (commented):
Reviewd.

Xierumeng (approved):
Approved.

PR #195, opened 2024-08-23
upgrade to python 3.11

maxlou05:
I didn't see any modules that directly utilized pymavlink or dronekit, so should be fine.


TongguangZhang (approved):
lgtm

PR #209, opened 2024-10-09
redo of integrating cluster estimation worker into main

Aleksa-M:


main_2024.py
```diff
@@ -374,7 +412,16 @@ def main() -> int:
                 main_logger.debug(
                     "geolocation confidence: " + str(detection_world.confidence), True
                 )
```
mgupta27, line 414: You can get rid of the geolocation logging to main now. Geolocation already has its own logging.

main_2024.py
```diff
@@ -374,7 +412,16 @@ def main() -> int:
...
-
+        try:
+            cluster_estimations = cluster_estimation_to_main_queue.queue.get_nowait()
```
mgupta27, line 416: maybe rename to `cluster_estimation_data`.

main_2024.py
```diff
@@ -374,7 +412,16 @@ def main() -> int:
...
+        except queue.Empty:
+            cluster_estimations = None
+        if cluster_estimations is not None:
```
mgupta27, line 419: Add a space above this line.

main_2024.py
```diff
@@ -374,7 +412,16 @@ def main() -> int:
...
+            for cluster in cluster_estimations:
```
mgupta27, line 420: `for object_in_world in ... `

main_2024.py
```diff
@@ -374,7 +412,16 @@ def main() -> int:
...
+                main_logger.debug("Cluser in world: True")
```
mgupta27, line 421: Fix spelling mistake with cluster. Also `True` is a parameter, not a part of the string literal.

main_2024.py
```diff
@@ -374,7 +412,16 @@ def main() -> int:
...
+                main_logger.debug("Cluster location x: " + str(cluster.location_x))
+                main_logger.debug("Cluster location y: " + str(cluster.location_y))
+                main_logger.debug("Cluster spherical variance: " + str(cluster.spherical_variance))
```
mgupta27, line 424: Rather than parsing the fields of the object here, I think its better to override the `__str__` method for that class. You can see `detection_in_world` as an example and do the same for `object_in_world`. Then in here you can just log the object in one line. To see an example of that look at `geolocation.py line 320`.

mgupta27 (commented):
Reviewed.

main_2024.py
```diff
@@ -357,24 +395,14 @@ def main() -> int:
...
-                main_logger.debug(
-                    "geolocation confidence: " + str(detection_world.confidence), True
-                )
-
+            cluster_estimation_data = None
+        
+        if cluster_estimation_data is not None:
+            for object_in_world in cluster_estimation_data:
+                main_logger.debug("Cluster in world: ", True)
+                main_logger.debug(object_in_world.__str__)
```
mgupta27, line 405: Instead of doing this to call the to string method you can just do `str(object_in_world)`.

main_2024.py
```diff
@@ -111,6 +112,10 @@ def main() -> int:
         GEOLOCATION_CAMERA_ORIENTATION_YAW = config["geolocation"]["camera_orientation_yaw"]
         GEOLOCATION_CAMERA_ORIENTATION_PITCH = config["geolocation"]["camera_orientation_pitch"]
         GEOLOCATION_CAMERA_ORIENTATION_ROLL = config["geolocation"]["camera_orientation_roll"]
+
+        MIN_ACTIVATION_THRESHOLD = config["cluster_merge"]["min_activation_threshold"]
+        MIN_NEW_POINTS_TO_RUN = config["cluster_merge"]["min_new_points_to_run"]
+        RANDOM_STATE = config["cluster_merge"]["random_state"]
```
mgupta27, line 118: I think you forgot to change the name from `cluster_merge` to `cluster_estimation` here. Can you try running main as well so that we don't miss errors like these as well. If you need help with that ping me and let me know.

main_2024.py
mgupta27, line 405: There should be a parameter set to `True` here. Please add that so we log the data to the file.

mgupta27 (commented):
Reviewed. Linters are failing. Please run `black --check .` to see if the code looks all good and `black .` to fix any files that cause an issue.

main_2024.py
```diff
@@ -357,24 +395,14 @@ def main() -> int:
...
-                main_logger.debug("geolocation label: " + str(detection_world.label), True)
-                main_logger.debug(
-                    "geolocation confidence: " + str(detection_world.confidence), True
-                )
+            cluster_estimation_data = None
 
+        if cluster_estimation_data is not None:
+            for object_in_world in cluster_estimation_data:
+                main_logger.debug("Cluster in world: ", (object_in_world is not None))
+                main_logger.debug(str(object_in_world))
```
mgupta27, line 405: While its good to check `object_in_world` is not `None`, I don't think its needed given that in the `ClusterEstimation.run()` it already checks that there are no `None` values in the list. Replace `(object_in_world is not None)` with `True`.

mgupta27 (commented):
Reviewed.

mgupta27 (approved):
Approved!

PR #217, opened 2024-11-03
Added logs to detect target

wdan31:
The logs describe time it took to process images and when images were processed.


modules/detect_target/detect_target_ultralytics.py
```diff
@@ -108,4 +113,8 @@ def run(
         if self.__show_annotations:
             cv2.imshow("Annotated", image_annotated)  # type: ignore
 
+        end_time = time.time()
+        self.__logger.info(
+            f"{time.localtime()}: Target detection took {round(end_time - start_time, 3)} seconds"
```
maxlou05, line 118: For time, it's fine to just use time.time() (logger automatically puts a timestamp with a proper format like hh:mm:ss), and there's no need to round it since we would rather have more precision. (these tasks usually take much less than 1 second, so more precision is required)

maxlou05 (commented):
Also, please update the common submodule (easiest way is to just run `setup_project.ps1` or `source setup_project.sh`)

maxlou05 (approved):
Approved!

modules/detect_target/detect_target_ultralytics.py
```diff
@@ -91,15 +96,15 @@ def run(
 
                 detections.append(detection)
 
+        end_time = time.time()
+        self.__local_logger.info(
+            f"{time.time()}: Count: {self.__counter}. Target detection took {end_time - start_time} seconds. Objects detected: {repr(detections)}."
```
Xierumeng, line 101: Remove `repr()` , only need `{detections}` .

modules/detection_in_world.py
```diff
@@ -53,6 +53,12 @@ def __init__(
         self.label = label
         self.confidence = confidence
 
+    def __repr__(self) -> str:
+        """
+        For collections (e.g. list).
+        """
+        return str(self)
```
Xierumeng, line 60: Move below `__str__()` .

modules/detections_and_time.py
```diff
@@ -57,6 +57,12 @@ def __str__(self) -> str:
         """
         return f"cls: {self.label}, conf: {self.confidence}, bounds: {self.x_1} {self.y_1} {self.x_2} {self.y_2}"
 
+    def __repr__(self) -> str:
+        """
+        For collections (e.g. list).
+        """
+        return str(self)
```
Xierumeng, line 64: Move below `__str__()` .

tests/unit/test_detect_target_ultralytics.py
```diff
@@ -108,8 +109,12 @@ def detector() -> detect_target_ultralytics.DetectTargetUltralytics:  # type: ig
     """
     Construct DetectTargetUltralytics.
     """
+    _, test_logger = logger.Logger.create("test_logger", False)
```
Xierumeng, line 112: `result, test_logger = ...` and assert result is true.

PR #237, opened 2025-01-16
Restart worker patch with fixed commit history

Adityya-K:
This is a new PR for the restart workers that fixes the messy commit history and implements some new changes.

Some things to note:
- I have only commented out the code that drains the queue just in case it is needed for the program to run
- Additionally, I don't quite understand how the queues are added to the new workers or if anything needs to be done to add them, so, if there are glaring flaws, I apologize

### DIFF
--- a/modules/communications/communications_worker.py
+++ b/modules/communications/communications_worker.py
@@ -7,7 +7,6 @@ import pathlib
 import queue
 
 from modules import object_in_world
-
 from . import communications
 from utilities.workers import queue_proxy_wrapper
 from utilities.workers import worker_controller
@@ -78,7 +77,7 @@ def communications_worker(
         if is_invalid:
             continue
 
-        result, value = comm.run(input_queue.queue.get())
+        result, value = comm.run(input_data)
         if not result:
             continue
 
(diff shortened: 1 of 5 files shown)


modules/communications/communications_worker.py
```diff
@@ -59,6 +61,23 @@ def communications_worker(
...
+        for single_input in input_data:
+            if not isinstance(single_input, object_in_world.ObjectInWorld):
+                local_logger.warning(f"Skipping unexpected input: {input}")
+                is_invalid = True
+                break
+
+        if is_invalid:
+            continue
+
         result, value = comm.run(input_queue.queue.get())
```
HermanG05, line 81: Replace the argument with input_data

modules/communications/communications_worker.py
```diff
@@ -6,6 +6,8 @@
 import pathlib
 import queue
 
+from modules import object_in_world
+
```
HermanG05, line 10: Remove space

HermanG05 (approved):
Conditionally Approved. Just address the two comments and push your changes before merging.

PR #242, opened 2025-01-22
Add configs to detect target pipeline

siddhp1:


config.yaml
```diff
@@ -22,10 +22,12 @@ video_input:
...
-    option: 0  # 0 is for Ultralytics (from detect_target_factory.py)
+    option: 0  # 0 is for Ultralytics and 1 is for brightspot (from detect_target_factory.py)
+    save_prefix: "log_comp"
+
+detect_ultralytics:
     device: 0
     model_path: "tests/model_example/yolov8s_ultralytics_pretrained_default.pt"  # See autonomy OneDrive for latest model
-    save_prefix: "log_comp"
 
 detect_brightspot:
```
maxlou05, line 32: Sorry just noticed this now, can we move this section under `detect_target` and call it `config`, like it is for `video_input`? In this case, you will comment out this whole section because we are using the ultralytics config

config.yaml
```diff
@@ -22,10 +22,12 @@ video_input:
 
 detect_target:
     worker_count: 1
-    option: 0  # 0 is for Ultralytics (from detect_target_factory.py)
+    option: 0  # 0 is for Ultralytics and 1 is for brightspot (from detect_target_factory.py)
+    save_prefix: "log_comp"
+
+detect_ultralytics:
```
maxlou05, line 28: Can we move this under `detect_target` and call it `config`, like it is for `video_input`?

main_2024.py
```diff
@@ -105,17 +107,48 @@ def main() -> int:
...
+            "max_inertia_ratio"
+        ]
+        DETECT_TARGET_BRIGHTSPOT_FILTER_BY_CONVEXITY = config["detect_brightspot"][
+            "filter_by_convexity"
+        ]
+        DETECT_TARGET_BRIGHTSPOT_MIN_CONVEXITY = config["detect_brightspot"]["min_convexity"]
+        DETECT_TARGET_BRIGHTSPOT_MAX_CONVEXITY = config["detect_brightspot"]["max_convexity"]
+        DETECT_TARGET_BRIGHTSPOT_FILTER_BY_AREA = config["detect_brightspot"]["filter_by_area"]
+        DETECT_TARGET_BRIGHTSPOT_MIN_AREA_PIXELS = config["detect_brightspot"]["min_area_pixels"]
+        DETECT_TARGET_BRIGHTSPOT_MAX_AREA_PIXELS = config["detect_brightspot"]["max_area_pixels"]
```
maxlou05, line 151: Instead of having this many variables, just have one `DETECT_TARGET_CONFIG` and use match case here (see `VIDEO_INPUT_CAMERA_CONFIG` above

modules/detect_target/detect_target_factory.py
```diff
@@ -21,9 +21,10 @@ class DetectTargetOption(enum.Enum):
...
-    device: "str | int",
-    model_path: str,
-    override_full: bool,
+    config: (
+        detect_target_brightspot.DetectTargetBrightspotConfig
+        | detect_target_ultralytics.DetectTargetUltralyticsConfig
+    ),
     local_logger: logger.Logger,
     show_annotations: bool,
     save_name: str,
```
maxlou05, line 30: Can we re-order these so that it goes `save_name`, `show_annotations`, `config`, `local_logger`?

maxlou05 (commented):
Reviewed

maxlou05 (approved):
LGTM (as long as you have sort of tested it?)

PR #243, opened 2025-01-22
Restart workers patch with fixes

Adityya-K:
Another pr because the previous pr had some errors that need to be addressed (sorry about that!)

### DIFF
--- a/utilities/workers/worker_manager.py
+++ b/utilities/workers/worker_manager.py
@@ -244,11 +244,4 @@ class WorkerManager:
 
         self.__workers = new_workers
 
-        # Draining the preceding queue ensures that the preceding queue data wasn't what
-        # caused the worker to fail. Draining the succeeding queues is not needed
-        # because a worker that died would not have put bad data into the queue.
-        # input_queues = self.__worker_properties.get_input_queues()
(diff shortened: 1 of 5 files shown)


utilities/workers/worker_manager.py
```diff
@@ -244,11 +244,4 @@ def check_and_restart_dead_workers(self) -> bool:
 
         self.__workers = new_workers
 
-        # Draining the preceding queue ensures that the preceding queue data wasn't what
-        # caused the worker to fail. Draining the succeeding queues is not needed
-        # because a worker that died would not have put bad data into the queue.
-        # input_queues = self.__worker_properties.get_input_queues()
-        # for queue in input_queues:
-        #     queue.drain_queue()
```
HermanG05, line 252: You can just remove these comments completely

HermanG05 (changes requested):
Reviewed

utilities/workers/worker_manager.py
Adityya-K: Yes they are deleted in the latest version!

HermanG05 (approved):
LGTM

PR #245, opened 2025-01-24
Using message_encode_decode to send MAVlink messages

HermanG05:
Used the following functions in data_encoding and flight_controller.py to send MAVlink messages to the flight controller via the flight interface worker:
- encode_metadata()
- encode_position_global()
- send_statustext_msg()

### DIFF
--- a/modules/communications/communications_worker.py
+++ b/modules/communications/communications_worker.py
@@ -10,6 +10,7 @@ from modules import object_in_world
 from . import communications
 from utilities.workers import queue_proxy_wrapper
 from utilities.workers import worker_controller
+from ..common.modules.data_encoding import metadata_encoding_decoding
 from ..common.modules.logger import logger
 
 
@@ -49,7 +50,7 @@ def communications_worker(
 
     local_logger.info(f"Home position received: {home_position}", True)
 
-    result, comm = communications.Communications.create(home_position, local_logger)
+    result, comm = communications.Communications.create(home_position, local_logger, worker_name)
     if not result:
         local_logger.error("Worker failed to create class object", True)
         return
@@ -79,8 +80,18 @@ def communications_worker(
         if is_invalid:
             continue
 
-        result, value = comm.run(input_data)
+        result, list_of_messages = comm.run(input_data)
         if not result:
             continue
 
-        output_queue.queue.put(value)
+        result, metadata = metadata_encoding_decoding.encode_metadata(
+            f"{worker_name}", len(list_of_messages)
+        )
+        if not result:
+            local_logger.error("Failed to encode metadata", True)
+            continue
+
+        output_queue.queue.put(metadata)
+
+        for message in list_of_messages:
+            output_queue.queue.put(message)
--- a/modules/communications/communications_worker.py
+++ b/modules/communications/communications_worker.py
@@ -70,7 +70,9 @@ def communications_worker(
 
         for single_input in input_data:
             if not isinstance(single_input, object_in_world.ObjectInWorld):
-                local_logger.warning(f"Skipping unexpected input: {input}")
+                local_logger.warning(
+                    f"Skipping unexpected input: {input_data}, because of unexpected value: {single_input}"
+                )
                 is_invalid = True
                 break
 
--- a/modules/communications/communications_worker.py
+++ b/modules/communications/communications_worker.py
@@ -6,6 +6,7 @@ import os
 import pathlib
 import queue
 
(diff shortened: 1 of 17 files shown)


modules/flight_interface/flight_interface_worker.py
```diff
@@ -57,11 +64,45 @@ def flight_interface_worker(
...
+
+    result, metadata = metadata_encoding_decoding.encode_metadata(
+        f"{worker_name}", coordinates_input_queue.queue.qsize()
+    )
+    if not result:
+        local_logger.error("Failed to encode metadata", True)
+        return
+
+    message_flight_controller.send_statustext_msg(f"{metadata}")
+
```
maxlou05, line 82: We cannot create another instance of `FlightController` here, you need to add an input in the `FlightInterface.run()` method, input the list, and then do this in the `FlightInterface` class' `run()` which has a `FlightController` already.

modules/flight_interface/flight_interface_worker.py
```diff
@@ -57,11 +64,45 @@ def flight_interface_worker(
...
+        result, message = message_encoding_decoding.encode_position_global(
+            f"{worker_name}", coordinate
+        )
+        if not result:
+            local_logger.error("Failed to encode PositionGlobal object", True)
+            continue
+
+        message_flight_controller.send_statustext_msg(f"{message}")
+
```
maxlou05, line 105: The way this works is `communications` will send a list, and you need to edit the last line in `communcations.py` to return `objects_in_world_global` rather than the same thing it got as input. This list will be the input into `FlightInterface`. Then, the metadata message is on how long this list is, and then you send each element of this list (GPS coords) one by one.

maxlou05 (commented):
Reviewed

modules/communications/communications_worker.py
```diff
@@ -79,8 +80,18 @@ def communications_worker(
...
             continue
 
-        result, value = comm.run(input_data)
+        result, list_of_messages = comm.run(input_data)
         if not result:
             continue
 
-        output_queue.queue.put(value)
+        result, metadata = metadata_encoding_decoding.encode_metadata(
+            f"{worker_name}", len(list_of_messages)
```
maxlou05, line 88: Instead of using a `worker_name` input, just hard code "communications_worker" here. It's my bad that the common's function doesn't take in the enum as an input, I forgot about that part. We'll change that later, or you can do that right now?

modules/flight_interface/flight_interface.py
```diff
@@ -103,6 +103,11 @@ def run(self) -> "tuple[bool, odometry_and_time.OdometryAndTime | None]":
 
         self.__logger.info(str(odometry_and_time_object), True)
 
+        result = self.controller.send_statustext_msg(message)
+        if not result:
+            self.__logger.error("Failed to send statustext message", True)
+            return False, None
```
maxlou05, line 109: Don't need to fail here, just log error. This will make data merge think we didn't get a odometry object

modules/flight_interface/flight_interface_worker.py
```diff
@@ -62,7 +65,16 @@ def flight_interface_worker(
 
         time.sleep(period)
 
-        result, value = interface.run()
+        coordinate = coordinates_input_queue.queue.get()
```
maxlou05, line 68: I think there is a `get_no_wait()`, and then if it's `None`, means queue is empty. If queue is empty, skip getting and just run with inputting `None` for coordinate. Then, flight interface will be able to run without stalling for communications worker.

modules/flight_interface/flight_interface_worker.py
```diff
@@ -62,7 +65,16 @@ def flight_interface_worker(
...
+        if coordinate is None:
+            local_logger.info("Received type None, exiting")
+            break
```
maxlou05, line 71: Don't exit, just get odometry only (see above) since we'll need to get many odometries before we can send 1 message

modules/communications/communications_worker.py
```diff
@@ -79,8 +80,18 @@ def communications_worker(
...
+        )
+        if not result:
+            local_logger.error("Failed to encode metadata", True)
+            continue
+
+        output_queue.queue.put(metadata)
```
maxlou05, line 94: Add a period (worker period) parameter, and wait that much time between placing things into queue (after this and in between the things in the loop)

maxlou05 (commented):
Reviewed

PR #246, opened 2025-01-25
integrate common changes for picamera library bugfix

maxlou05:
Also set CI to not run integration tests, and slightly updated setup_project to not deactivate


BalajiLeninrajan (approved):
lgtm

PR #247, opened 2025-01-29
Using message_encode_decode to send MAVlink messages

HermanG05:
Fixed PR with cleaner commit history.

Description:
Used the following functions in data_encoding and flight_controller.py to send MAVlink messages to the flight controller via the flight interface worker:

encode_metadata()
encode_position_global()
send_statustext_msg()

### DIFF
--- a/modules/communications/communications.py
+++ b/modules/communications/communications.py
@@ -8,6 +8,8 @@ from .. import object_in_world
 from ..common.modules import position_global
 from ..common.modules import position_local
 from ..common.modules.logger import logger
+from ..common.modules.data_encoding import message_encoding_decoding
+from ..common.modules.data_encoding import worker_enum
 from ..common.modules.mavlink import local_global_conversion
 
 
@@ -23,6 +25,7 @@ class Communications:
         cls,
         home_position: position_global.PositionGlobal,
         local_logger: logger.Logger,
+        worker_id: int,
     ) -> "tuple[True, Communications] | tuple[False, None]":
         """
         Logs data and forwards it.
@@ -32,13 +35,14 @@ class Communications:
         Returns: Success, class object.
         """
 
-        return True, Communications(cls.__create_key, home_position, local_logger)
+        return True, Communications(cls.__create_key, home_position, local_logger, worker_id)
 
     def __init__(
         self,
         class_private_create_key: object,
         home_position: position_global.PositionGlobal,
         local_logger: logger.Logger,
+        worker_id: int,
     ) -> None:
         """
         Private constructor, use create() method.
@@ -47,11 +51,12 @@ class Communications:
 
         self.__home_position = home_position
         self.__logger = local_logger
+        self.__worker_id = worker_id
 
     def run(
         self,
         objects_in_world: list[object_in_world.ObjectInWorld],
-    ) -> tuple[True, list[object_in_world.ObjectInWorld]] | tuple[False, None]:
+    ) -> tuple[True, list[bytes]] | tuple[False, None]:
 
         objects_in_world_global = []
         for object_in_world in objects_in_world:
@@ -87,4 +92,16 @@ class Communications:
 
         self.__logger.info(f"{time.time()}: {objects_in_world_global}")
 
-        return True, objects_in_world
+        encoded_position_global_objects = []
+        for object in object_in_world_global:
+
+            result, message = message_encoding_decoding.encode_position_global(
+                self.__worker_id, object
+            )
+            if not result:
+                self.__logger.warning("Conversion from PositionGlobal to bytes failed", True)
+                return False, None
+
+            encoded_position_global_objects.append(message)
+
+        return True, encoded_position_global_objects
--- a/modules/communications/communications_worker.py
+++ b/modules/communications/communications_worker.py
@@ -5,16 +5,20 @@ Logs data and forwards it.
 import os
 import pathlib
 import queue
+import time
 
 from modules import object_in_world
 from . import communications
 from utilities.workers import queue_proxy_wrapper
 from utilities.workers import worker_controller
+from ..common.modules.data_encoding import metadata_encoding_decoding
+from ..common.modules.data_encoding import worker_enum
 from ..common.modules.logger import logger
 
 
 def communications_worker(
     timeout: float,
+    period: float,
     home_position_queue: queue_proxy_wrapper.QueueProxyWrapper,
     input_queue: queue_proxy_wrapper.QueueProxyWrapper,
     output_queue: queue_proxy_wrapper.QueueProxyWrapper,
@@ -29,6 +33,7 @@ def communications_worker(
     """
 
     worker_name = pathlib.Path(__file__).stem
+    worker_id = worker_enum.WorkerEnum.COMMUNICATIONS_WORKER
     process_id = os.getpid()
     result, local_logger = logger.Logger.create(f"{worker_name}_{process_id}", True)
     if not result:
@@ -49,7 +54,7 @@ def communications_worker(
 
     local_logger.info(f"Home position received: {home_position}", True)
 
-    result, comm = communications.Communications.create(home_position, local_logger)
+    result, comm = communications.Communications.create(home_position, local_logger, worker_id)
     if not result:
         local_logger.error("Worker failed to create class object", True)
         return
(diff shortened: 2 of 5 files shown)


modules/common
maxlou05, line 1: Where is this change? Can you make this a PR in common or something?

modules/communications/communications_worker.py
```diff
@@ -5,16 +5,20 @@
...
+from ..common.modules.data_encoding import worker_enum
 from ..common.modules.logger import logger
 
 
 def communications_worker(
     timeout: float,
+    period: float,
     home_position_queue: queue_proxy_wrapper.QueueProxyWrapper,
     input_queue: queue_proxy_wrapper.QueueProxyWrapper,
     output_queue: queue_proxy_wrapper.QueueProxyWrapper,
```
maxlou05, line 24: There should be another output queue here (to flight controller), unless you change the output queue (which is currently to main) to be the flight controller queue

modules/communications/communications.py
```diff
@@ -23,6 +25,7 @@ def create(
         cls,
         home_position: position_global.PositionGlobal,
         local_logger: logger.Logger,
+        worker_id: worker_enum.WorkerEnum,
```
maxlou05, line 28: Instead of having it as an input, the communications worker/class should always use the communication worker id. Just use the enum when you need to as a constant instead of as an input

modules/communications/communications_worker.py
```diff
@@ -79,8 +84,21 @@ def communications_worker(
...
 
-        output_queue.queue.put(value)
+        result, metadata = metadata_encoding_decoding.encode_metadata(
+            worker_id, len(list_of_messages)
+        )
+        if not result:
+            local_logger.error("Failed to encode metadata", True)
+            continue
+
+        output_queue.queue.put(metadata)
```
maxlou05, line 98: use the flight controller queue (maybe call it message queue?) for these. Then you can also output it to main queue as well for logging/debugging (so just output everything to both queues)

modules/flight_interface/flight_interface.py
```diff
@@ -103,6 +103,10 @@ def run(self) -> "tuple[bool, odometry_and_time.OdometryAndTime | None]":
 
         self.__logger.info(str(odometry_and_time_object), True)
 
+        result = self.controller.send_statustext_msg(message)
+        if not result:
+            self.__logger.error("Failed to send statustext message", True)
+
```
maxlou05, line 109: Note that `message` may be `None` due to `get_nowait`, so you should not log error every time. Skip normally if its `None`

modules/flight_interface/flight_interface_worker.py
```diff
@@ -20,6 +21,7 @@ def flight_interface_worker(
     input_queue: queue_proxy_wrapper.QueueProxyWrapper,
     output_queue: queue_proxy_wrapper.QueueProxyWrapper,
     communications_output_queue: queue_proxy_wrapper.QueueProxyWrapper,
+    coordinates_input_queue: queue_proxy_wrapper.QueueProxyWrapper,
```
maxlou05, line 24: This must be the second queue becasue that's what you defined in main_2024. It goes in order (input1, input2, output1, output2...)

modules/flight_interface/flight_interface_worker.py
```diff
@@ -62,7 +64,17 @@ def flight_interface_worker(
 
         time.sleep(period)
 
-        result, value = interface.run()
+        try:
+            coordinate = coordinates_input_queue.queue.get_nowait()
+        except queue.Empty:
+            local_logger.warning("No more coordinates to process")
```
maxlou05, line 70: No need to log this, as it will log it 8 times per second, and it's normal operation

modules/flight_interface/flight_interface_worker.py
```diff
@@ -62,7 +64,17 @@ def flight_interface_worker(
...
+            coordinate = None
+
+        if not isinstance(coordinate, bytes):
+            local_logger.warning(f"Skipping unexpected input: {coordinate}")
+            continue
```
maxlou05, line 75: Do not put this check here, put it in the run() function. This is so that it still continues to put odometry in the output queues, and you don't skip that functionality of the flight interface worker (cuz now you'll just skip getting odometry every time you don't have a communications message)

modules/flight_interface/flight_interface_worker.py
```diff
@@ -4,6 +4,7 @@
 
 import os
 import pathlib
+import queue
```
maxlou05, line 7: Is it necessary to import queue here? I don't really know how this works, but it feels like it's not needed?

tests/integration/test_flight_interface_worker.py
```diff
@@ -11,11 +11,34 @@
...
 
 MAVLINK_CONNECTION_ADDRESS = "tcp:localhost:14550"
 FLIGHT_INTERFACE_TIMEOUT = 10.0  # seconds
 FLIGHT_INTERFACE_BAUD_RATE = 57600  # symbol rate
 FLIGHT_INTERFACE_WORKER_PERIOD = 0.1  # seconds
+WORK_COUNT = 4
+COMMUNICATIONS_WORKER_ID = worker_enum.WorkerEnum.COMMUNICATIONS_WORKER
+
+
+def simulate_communications_worker(
```
maxlou05, line 26: No need to add this test, as Jane is doing it as her task. Just make sure that the old test can still run (ie fix the inputs to the functions and worker)

maxlou05 (commented):
Reviewed

maxlou05 (commented):
Reviewed. Conditionally approved after fix and common PR

modules/communications/communications_worker.py
```diff
@@ -79,8 +84,23 @@ def communications_worker(
...
         if not result:
             continue
 
-        output_queue.queue.put(value)
+        result, metadata = metadata_encoding_decoding.encode_metadata(
+            worker_enum.WorkerEnum.COMMUNICATIONS_WORKER, len(list_of_messages)
+        )
+        if not result:
+            local_logger.error("Failed to encode metadata", True)
+            continue
```
maxlou05, line 96: Could you move this thing to `communications.py`? Then this file will no longer need to import enum and encoding, and just do sending.

modules/flight_interface/flight_interface.py
```diff
@@ -103,6 +103,8 @@ def run(self) -> "tuple[bool, odometry_and_time.OdometryAndTime | None]":
 
         self.__logger.info(str(odometry_and_time_object), True)
 
+        self.controller.send_statustext_msg(message)
```
maxlou05, line 106: Check if message is `None`. If it's `None`, do not send any message and just continue normal operation

modules/flight_interface/flight_interface.py
```diff
@@ -73,7 +73,7 @@ def get_home_position(self) -> position_global.PositionGlobal:
         """
         return self.__home_position
 
-    def run(self) -> "tuple[bool, odometry_and_time.OdometryAndTime | None]":
+    def run(self, message: bytes) -> "tuple[bool, odometry_and_time.OdometryAndTime | None]":
```
maxlou05, line 76: `message: bytes | None`

maxlou05 (commented):
Reviewed

modules/communications/communications.py
```diff
@@ -81,10 +84,29 @@ def run(
...
+            encoded_position_global_objects.append(message)
+
+        result, metadata = metadata_encoding_decoding.encode_metadata(
+            worker_enum.WorkerEnum.COMMUNICATIONS_WORKER, len(encoded_position_global_objects)
+        )
+        if not result:
+            self.__logger.error("Failed to encode metadata", True)
+            return False, None, None
+
+        return True, encoded_position_global_objects, metadata
```
maxlou05, line 112: Just switch the order: metadata then encoded_position_global_objects. Make sure to change function definition and also communications worker.

modules/communications/communications_worker.py
```diff
@@ -79,8 +82,16 @@ def communications_worker(
...
-        result, value = comm.run(input_data)
+        result, list_of_messages, metadata = comm.run(input_data)
         if not result:
             continue
 
-        output_queue.queue.put(value)
+        output_queue.queue.put(metadata)
+        message_output_queue.queue.put(metadata)
+
+        for message in list_of_messages:
```
maxlou05, line 92: Actually, it seems that your run() will return None, so this may throw exception. Either return a list (ie `[None]`) or do a check here (it might be more appropriate to do a check here)

maxlou05 (commented):
Reviewed

maxlou05 (approved):
Approved

PR #251, opened 2025-02-07
Cluster estimation integration test

RohanKatreddy:


tests/integration/test_cluster_estimation_worker.py
```diff
@@ -0,0 +1,142 @@
...
+Test cluster_estimation_worker process.
+"""
+
+import time
+import multiprocessing as mp
+from typing import List
+import numpy as np
+from utilities.workers import queue_proxy_wrapper, worker_controller
+from modules.detection_in_world import DetectionInWorld
+from modules.cluster_estimation.cluster_estimation_worker import cluster_estimation_worker
```
HermanG05, line 11: Ensure that your imports follow the styling convention: https://uwarg-docs.atlassian.net/wiki/spaces/CV/pages/2253226033/Python+Style+Convention

tests/integration/test_cluster_estimation_worker.py
```diff
@@ -0,0 +1,142 @@
...
+    input_queue = queue_proxy_wrapper.QueueProxyWrapper(mp_manager)
+    output_queue = queue_proxy_wrapper.QueueProxyWrapper(mp_manager)
+
+    worker_process = mp.Process(
+        target=cluster_estimation_worker,
+        args=(
+            3,
+            0,
+            3,
+            0,
```
HermanG05, line 32: Try to store numerical constants like these at the top (ie RANDOM_STATE = 0, see other integration tests as an example).

tests/integration/test_cluster_estimation_worker.py
```diff
@@ -0,0 +1,142 @@
...
+        )[1],
+    ]
+
+    # Testing with test_data_1 (1 cluster)
+
+    input_queue.queue.put(test_data_1)
+    worker_process.start()
+    time.sleep(1)
+
+    output_results: List[List[DetectionInWorld]] = output_queue.queue.get()
```
HermanG05, line 113: We want to check to see if the output result is a list of ObjectInWorld, otherwise the worker wouldnt be doing anything

tests/integration/test_cluster_estimation_worker.py
```diff
@@ -0,0 +1,142 @@
...
+    assert len(output_results) == 1
+
+    time.sleep(1)
+
+    # Testing with test_data_2 (2 clusters)
+
+    input_queue.queue.put(test_data_2)
+    time.sleep(1)
+
+    output_results: List[List[DetectionInWorld]] = output_queue.queue.get()
```
HermanG05, line 125: same here

HermanG05 (changes requested):
Reviewed

tests/integration/test_cluster_estimation_worker.py
```diff
@@ -0,0 +1,154 @@
...
+    input_queue.queue.put(test_data_1)
+    worker_process.start()
+    time.sleep(1)
+
+    output_results: List[List[DetectionInWorld]] = output_queue.queue.get()
+
+    assert output_results is not None
+    assert isinstance(output_results, list)
+    assert len(output_results) == 1
+    assert all(isinstance(obj, ObjectInWorld) for obj in output_results)
```
HermanG05, line 126: Sorry maybe I should've been more specific. On line 121 and 135 where you're using the .get() method to remove a list from the queue, you want the variable to expect a List[ObjectInWorld], not a List[List[DetectionInWorld]]. The get method returns a single element from the queue, and in this case, each element should be a List[ObjectInWorld].

HermanG05 (commented):
Reviewed.

tests/integration/test_cluster_estimation_worker.py
RohanKatreddy: ah ok mb, simple fix then

tests/integration/test_cluster_estimation_worker.py
```diff
@@ -0,0 +1,154 @@
...
+
+    input_queue.queue.put(test_data_2)
+    time.sleep(1)
+
+    output_results: List[DetectionInWorld] = output_queue.queue.get()
+
+    assert output_results is not None
+    assert isinstance(output_results, list)
+    assert len(output_results) == 2
+    assert all(isinstance(obj, ObjectInWorld) for obj in output_results)
```
HermanG05, line 140: It might be better to create a loop where you extract each list from the queue and check whether they are the types youre expecting. You'd remove the len(output_results) assertion and just wrap the other ones inside the loop, exiting if they fail. I would probably try to do this with test_data_1 as well.

HermanG05 (commented):
Reviewed

tests/integration/test_cluster_estimation_worker.py
```diff
@@ -0,0 +1,157 @@
...
+
+def check_output_results(output_queue: queue_proxy_wrapper.QueueProxyWrapper) -> None:
+    """
+    Checking if the output from the worker is of the correct type
+    """
+
+    output_results: List[DetectionInWorld] = output_queue.queue.get()
+
+    while not output_queue.queue.empty():
+        output_results = output_queue.queue.get()
```
HermanG05, line 30: Move output_results: List[DetectionInWorld] = output_queue.queue.get() to line 30 and remove it from line 27.

HermanG05 (approved):
Conditionally Approved

PR #253, opened 2025-02-14
Cluster estimation filter

szbillyang:


modules/cluster_estimation/cluster_estimation.py
```diff
@@ -337,15 +348,16 @@ def __filter_by_points_ownership(
         # List of each point's cluster index
         cluster_assignment = self.__vgmm.predict(self.__all_points)  # type: ignore
 
-        # Find which cluster indices have points
-        clusters_with_points = np.unique(cluster_assignment)
+        # Get counts for each cluster index
+        unique, counts = np.unique(cluster_assignment, return_counts=True)
+        cluster_counts = dict(zip(unique, counts))
 
         # Remove empty clusters
```
maxlou05, line 355: Edit this comment to reflect your new code: Filter by number of points corresponding to each cluster
Also edit the docstring to reflect this change.

modules/cluster_estimation/cluster_estimation.py
```diff
@@ -337,15 +348,16 @@ def __filter_by_points_ownership(
...
         filtered_output: "list[tuple[np.ndarray, float, float]]" = []
         # By cluster index
```
maxlou05, line 357: Edit this to say: check each cluster has enough points associated to it by index

modules/cluster_estimation/cluster_estimation.py
```diff
@@ -337,15 +348,16 @@ def __filter_by_points_ownership(
...
         # pylint: disable-next=consider-using-enumerate
         for i in range(len(model_output)):
-            if i in clusters_with_points:
+            if cluster_counts.get(i, 0) >= self.__min_points_per_cluster:
```
maxlou05, line 360: Why is it necessary to put another 0? Is it not just `.get(i)`?

tests/unit/test_cluster_detection.py
```diff
@@ -37,6 +38,7 @@ def cluster_model() -> cluster_estimation.ClusterEstimation:  # type: ignore
         MAX_NUM_COMPONENTS,
         RNG_SEED,
         test_logger,
+        MIN_POINTS_PER_CLUSTER,
```
maxlou05, line 41: Did you verify the test cases to make sure that all the clusters have at least 3 points? Can you also make a test where there is a cluster that only has like 1 point (ie make a very far outlier)

maxlou05: Also, did you check the integration tests? They are under `tests/integration`

maxlou05 (commented):
Reviewed

maxlou05 (approved):
Approved!

PR #254, opened 2025-02-14
Min brightness filter

vichua2006:


modules/detect_target/detect_target_brightspot.py
```diff
@@ -132,13 +135,18 @@ def run(
             )
             return False, None
 
+        # Apply min brightness thresholding first to remove non-bright-spot noise
+        threshold_used, min_filtered_image = cv2.threshold(
+            grey_image, self.__config.min_brightness_threshold, 255, cv2.THRESH_TOZERO
+        )
+
```
HermanG05, line 142: It might be a good idea to add a check after this to see whether the method was successful before continuing

HermanG05 (commented):
Reviewed

modules/detect_target/detect_target_brightspot.py
maxlou05: Just a quick question, why are you using `THRESH_TOZERO` while the original one uses `THRESH_BINARY`? Is there a difference?

config.yaml
```diff
@@ -45,6 +45,7 @@ detect_target:
     #     filter_by_area: True
     #     min_area_pixels: 50
     #     max_area_pixels: 640
+    #     min_brightness_threshold: 50
```
maxlou05, line 48: please make this a lot higher, like at least 200, probably 225

maxlou05 (commented):
Reviewed

modules/detect_target/detect_target_brightspot.py
vichua2006: > It might be a good idea to add a check after this to see whether the method was successful before continuing

Roger that, will do the error check.

> Just a quick question, why are you using `THRESH_TOZERO` while the original one uses `THRESH_BINARY`? Is there a difference?

I was looking at [docs](https://docs.opencv.org/4.x/d7/d4d/tutorial_py_thresholding.html), and it seems that `THRESH_BINARY` sets all values greater than the threshold to a specified maximum value (255 in this case). I thought that would defeat the purpose of the percentile filter afterwards, since whatever passed the min brightness would be set to 255 and automatically pass the percentile filter as well, regardless of its original value. `THRESH_TOZERO` sets values lower than the threshold to 0, but keeps values above the threshold as they are, and allows them to be filtered again by the percentile filter.
image

but now that I'm looking at the filters again, maybe it doesn't matter too much if the percentile filter is set to 99.0.

config.yaml
vichua2006: > please make this a lot higher, like at least 200, probably 225

I originally had it at 200, but was failing the unit and the integration tests. It was able to detect the brightspots in the integration test at 100, and passed the unit test at 50. What should I do?

modules/detect_target/detect_target_brightspot.py
```diff
@@ -62,6 +64,7 @@ def __init__(
         filter_by_area: Whether to filter by area.
         min_area_pixels: Minimum area in pixels.
         max_area_pixels: Maximum area in pixels.
+        min_brightness_threshold: Minimum brightness threshold for bright spots.
```
maxlou05, line 67:
```
min_brightness_threshold: Minimum absolute brightness of individual pixels to be considered
min_average_brightness_threshold: Minimum absolute average brightness of detected blobs
```

modules/detect_target/detect_target_brightspot.py
```diff
@@ -132,18 +137,25 @@ def run(
...
+            grey_image, combined_threshold, 255, cv2.THRESH_BINARY
         )
         if threshold_used == 0:
-            self.__local_logger.error(f"{time.time()}: Failed to threshold image.")
+            self.__local_logger.error(f"{time.time()}: Failed to percentile threshold image.")
             return False, None
 
+        # cv2.imshow("Thresholded", bw_image)  # type: ignore
+        # cv2.waitKey(0)  # type: ignore
+
```
maxlou05, line 158: Delete

modules/detect_target/detect_target_brightspot.py
```diff
@@ -132,18 +137,25 @@ def run(
...
+        combined_threshold = max(brightspot_threshold, self.__config.min_brightness_threshold)
+
+        # Apply combined thresholding to isolate bright spots
         threshold_used, bw_image = cv2.threshold(
-            grey_image, brightspot_threshold, 255, cv2.THRESH_BINARY
+            grey_image, combined_threshold, 255, cv2.THRESH_BINARY
         )
         if threshold_used == 0:
-            self.__local_logger.error(f"{time.time()}: Failed to threshold image.")
+            self.__local_logger.error(f"{time.time()}: Failed to percentile threshold image.")
```
maxlou05, line 153: Don't print time.time(), just `"Failed to percentile threshold image"`

modules/detect_target/detect_target_brightspot.py
```diff
@@ -166,12 +178,57 @@ def run(
 
         # A lack of detections is not an error, but should still not be forwarded
         if len(keypoints) == 0:
-            self.__local_logger.info(f"{time.time()}: No brightspots detected.")
+            self.__local_logger.info(
+                f"{time.time()}: No brightspots detected (before blob average filter)."
```
maxlou05, line 182: Don't print time.time()

modules/detect_target/detect_target_brightspot.py
```diff
@@ -166,12 +178,57 @@ def run(
...
+        filtered_keypoints = []
+        for brightness, idx in average_brightness_list:
+            # Only append associated keypoint if the blob average is bright enough
+            if brightness >= self.__config.min_average_brightness_threshold:
+                filtered_keypoints.append(keypoints[idx])
+
+        # A lack of detections is not an error, but should still not be forwarded
+        if len(filtered_keypoints) == 0:
+            self.__local_logger.info(
+                f"{time.time()}: No brightspots detected (after blob average filter)."
```
maxlou05, line 225: Don't print time.time()

modules/detect_target/detect_target_brightspot.py
```diff
@@ -166,12 +178,57 @@ def run(
...
+            )
             return False, None
 
         # Annotate the image (green circle) with detected keypoints
         image_annotated = cv2.drawKeypoints(
-            image, keypoints, None, (0, 255, 0), cv2.DRAW_MATCHES_FLAGS_DRAW_RICH_KEYPOINTS
+            image, filtered_keypoints, None, (0, 255, 0), cv2.DRAW_MATCHES_FLAGS_DRAW_RICH_KEYPOINTS
         )
```
maxlou05, line 232: Can we move all of this inside of the if statement on line 269 (only annotate images if needed)

maxlou05 (commented):
Reviewed

maxlou05 (approved):
Approved for temporary Flight test integration integration

PR #262, opened 2025-07-18
Complete autonomous landing module

danielquzhao:
Completed the autonomous landing module. Tackled the following problems in the ticket:

Multiple detections? Which one do we choose to go to?

Implemented a flexible detection selection system with four different approaches:
• NEAREST_TO_CENTER (default): Chooses the detection closest to the image center
• LARGEST_AREA: Selects the detection with the biggest bounding box
• HIGHEST_CONFIDENCE: Picks the detection with the highest confidence score
• FIRST_DETECTION: Uses the first detection in the list

How/where does this get integrated in the worker pipeline?

The auto_landing_worker() function follows the worker pattern:
• Input: Receives merged_odometry_detections from an input queue
• Output: Sends landing_info to an output queue
• Commands: Receives control commands for state changes from a command queue

How do we turn it on/off? How do we signal this?

Uses AutoLandingCommand and AutoLandingController. The command wrapper carries two string commands ("enable"/"disable") and is sent through the command queue to be processed. The controller layer has enable() and disable() methods, and is_enabled() for status queries. The process_detections() only processes when enabled.

Let me know if anything needs changing.

Note: Some of the brightspot tests are failing so it doesn't pass the pytest check. I believe they were already failing beforehand.

### DIFF
--- a/modules/auto_landing/auto_landing.py
+++ b/modules/auto_landing/auto_landing.py
@@ -4,9 +4,23 @@ for use with LANDING_TARGET MAVLink command.
 """
 
 import math
+from enum import Enum
+from typing import Optional
+import threading
 
 from .. import merged_odometry_detections
 from ..common.modules.logger import logger
+
+
+class DetectionSelectionStrategy(Enum):
+    """
+    Strategies for selecting which detection to use when multiple targets are detected.
+    """
+
+    NEAREST_TO_CENTER = "nearest_to_center"  # Choose detection closest to image center
+    LARGEST_AREA = "largest_area"  # Choose detection with largest bounding box area
+    HIGHEST_CONFIDENCE = "highest_confidence"  # Choose detection with highest confidence
+    FIRST_DETECTION = "first_detection"  # Use first detection in list (original behavior)
 
 
 class AutoLandingInformation:
@@ -44,17 +58,21 @@ class AutoLanding:
         im_h: float,
         im_w: float,
         local_logger: logger.Logger,
+        selection_strategy: DetectionSelectionStrategy = DetectionSelectionStrategy.NEAREST_TO_CENTER,
     ) -> "tuple [bool, AutoLanding | None ]":
         """
         fov_x: The horizontal camera field of view in degrees.
         fov_y: The vertical camera field of view in degrees.
         im_w: Width of image.
         im_h: Height of image.
+        selection_strategy: Strategy for selecting which detection to use when multiple targets are detected.
 
         Returns an AutoLanding object.
         """
 
-        return True, AutoLanding(cls.__create_key, fov_x, fov_y, im_h, im_w, local_logger)
+        return True, AutoLanding(
+            cls.__create_key, fov_x, fov_y, im_h, im_w, local_logger, selection_strategy
+        )
 
     def __init__(
         self,
@@ -64,6 +82,7 @@ class AutoLanding:
         im_h: float,
         im_w: float,
         local_logger: logger.Logger,
+        selection_strategy: DetectionSelectionStrategy,
     ) -> None:
         """
         Private constructor, use create() method.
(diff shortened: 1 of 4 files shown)


modules/auto_landing/auto_landing.py
```diff
@@ -4,11 +4,25 @@
...
 from ..common.modules.logger import logger
 
 
+class DetectionSelectionStrategy(Enum):
+    """
+    Strategies for selecting which detection to use when multiple targets are detected.
+    """
+
+    NEAREST_TO_CENTER = "nearest_to_center"  # Choose detection closest to image center
+    LARGEST_AREA = "largest_area"  # Choose detection with largest bounding box area
```
ellyokes253, line 21: i don't think these two options will be relevant to our purposes 
- `NEAREST_TO_CENTER` won't matter as much because math can always be done to get the distance and location of the target
- `LARGEST_AREA` is also not as important because there could be a false detection in that list that is large. could be less likely with a properly trained model, but the other two imo seem better to do

modules/auto_landing/auto_landing.py
```diff
@@ -75,10 +94,75 @@ def __init__(
         self.im_h = im_h
         self.im_w = im_w
         self.__logger = local_logger
+        self.__selection_strategy = selection_strategy
+
+    def _select_detection(self, detections: list) -> Optional[int]:
```
ellyokes253, line 99: i think you don't have to include the `typing` class and can just specify return type without it

modules/auto_landing/auto_landing.py
```diff
@@ -75,10 +94,75 @@ def __init__(
...
+
+        Returns the index of the selected detection, or None if no suitable detection found.
+        """
+        if not detections:
+            return None
+
+        if len(detections) == 1:
+            return 0
+
+        if self.__selection_strategy == DetectionSelectionStrategy.FIRST_DETECTION:
```
ellyokes253, line 111: could put in the same if statement cause same output

modules/auto_landing/auto_landing_worker.py
```diff
@@ -42,25 +61,106 @@ def auto_landing_worker(
...
+        if input_data is not None:
+            result, landing_info = landing_controller.process_detections(input_data)
+            if result and landing_info:
+                output_queue.queue.put(landing_info)
 
-        output_queue.queue.put(value)
         time.sleep(period)
+
+
+def _process_commands(
```
ellyokes253, line 110: i think what you're trying to implement is a queue that has a bunch of commands meaning the autolander decides whether or not it wants to autoland, and the fact that it can change from wanting to or not during execution hence calling the function in the worker (correct me if i'm wrong). i think the simplistic way to go is to assume that we will know whether or not we want to autoland, and that decision will not change throughout the execution of the worker once it's decided.

modules/auto_landing/auto_landing_worker.py
```diff
@@ -12,22 +13,40 @@
 from ..common.modules.logger import logger
 
 
+class AutoLandingCommand:
+    """
+    Command structure for controlling auto-landing operations.
+    """
+
+    def __init__(self, command: str) -> None:
+        self.command = command
```
ellyokes253, line 22: what will make the most sense for making auto landing an enable/disable thing would be to include this as a constant in the `config.yaml` so that it's easily configurable

modules/auto_landing/auto_landing.py
```diff
@@ -108,3 +210,94 @@ def run(
...
+    ) -> None:
+        """
+        Private constructor, use create() method.
+        """
+        assert class_private_create_key is AutoLandingController.__create_key, "Use create() method"
+
+        self.__auto_landing = auto_landing
+        self.__logger = local_logger
+        self.__enabled = False
+        self.__enabled_lock = threading.Lock()
```
ellyokes253, line 252: by getting rid of the queue that could possibly introduce the race conditions of enabling and disabling auto landing, we won't need the thread lock anymore cause this can only change once

tests/integration/test_auto_landing_worker.py
```diff
@@ -0,0 +1,248 @@
...
+from utilities.workers import queue_proxy_wrapper
+from utilities.workers import worker_controller
+
+# Worker parameters
+FOV_X = 90.0  # degrees
+FOV_Y = 90.0  # degrees
+IMAGE_HEIGHT = 640.0  # pixels
+IMAGE_WIDTH = 640.0  # pixels
+WORKER_PERIOD = 0.1  # seconds
+DETECTION_STRATEGY = auto_landing.DetectionSelectionStrategy.NEAREST_TO_CENTER
```
ellyokes253, line 29: will need to change

tests/integration/test_auto_landing_worker.py
```diff
@@ -0,0 +1,268 @@
...
+# detection_strategy = auto_landing.DetectionSelectionStrategy.HIGHEST_CONFIDENCE
+
+# Ensure logs directory exists and create timestamped subdirectory
+LOG_DIR = pathlib.Path("logs")
+LOG_DIR.mkdir(parents=True, exist_ok=True)
+
+# Create a timestamped subdirectory for this test session
+TIMESTAMP_FORMAT = "%Y-%m-%d_%H-%M-%S"
+test_session_dir = LOG_DIR / datetime.datetime.now().strftime(TIMESTAMP_FORMAT)
+test_session_dir.mkdir(parents=True, exist_ok=True)
```
ellyokes253, line 40: i think we wont need to keep logs for the tests, the terminal print statements should suffice

tests/integration/test_auto_landing_worker.py
```diff
@@ -0,0 +1,268 @@
...
+    time.sleep(0.2)
+
+    # Should have output now
+    assert not output_queue.queue.empty()
+    landing_info = output_queue.queue.get_nowait()
+    assert landing_info is not None
+    assert hasattr(landing_info, "angle_x")
+    assert hasattr(landing_info, "angle_y")
+    assert hasattr(landing_info, "target_dist")
+    local_logger.info("--- Test 1 Passed ---")
```
ellyokes253, line 192: instead of using the logger, you can use print statements so these can be seen in the terminal while running tests

ellyokes253:
reviewed!

